Issue #120, opened 2013-12-03
Create json-stream reporter for Grunt task


pifantastic:
Currently, the intern grunt task checks the output of the reporter for `pas` or `fail` and uses `grunt.log.{ok|error}`. This is rather limiting as a reporter should have the freedom of using both `pass` or `fail` in its output language, regardless of the actual test result. Additionally, it's hard to know what chunk of data will be making it's way stdout at any one time.

Ideally, I think the intern grunt task would just pass stdout straight through, without altering the output. That probably means using `grunt.log.write` instead of `grunt.log.ok` or `grunt.log.error`. The `ok` and `error` methods also trim(), which can be annoying when trying to format things nicely.

jason0x43:
Just cleaning things up...

A better solution now would be for the grunt task to directly register listeners with the Intern executor. If that seems worthwhile, we can open a new issue along those lines.

PR #199, opened 2014-05-23
JSON stream reporter

jason0x43:
This PR adds a JSON stream reporter and updates the grunt task to use it. The reporter outputs runner status messages on stdout as JSON strings; all other output is redirected to stderr. Each JSON object contains the topic name, a default message, and the relevant object (test, suite, error).  See #120.

Documentation is in the `json-reporter` branch of the wiki repo.


jason0x43:
This should be rewritten after the [refactor-execution](https://github.com/theintern/intern/tree/refactor-execution) branch lands.

PR #309, opened 2014-11-21
Separate basePath from baseUrl

### LINKED ISSUE #249: Setting baseUrl breaks either node or browser client
Setting the `loader.baseUrl` config inherently breaks one of:

* node client - `my-project/node_modules/.bin/intern-client config=tests/intern`
* browser client `http://localhost/my-project/node_modules/intern/client.html?config=tests/intern`

~~due to its default value being set dynamically based on whether being run from node or the browser~~
due to the tests being ran from different path locations.

It seems natural to normalise this value with the place it is defined, i.e. the intern configuration file (tests/intern in this example), which would then allow a consistent baseUrl path to be applied across both node and browser clients.

The default baseUrl is already dynamically configured internally, allowing the user to statically set a baseUrl themselves will inherently break one of the two client side testing platforms.

There is a work-around - including the `intern/node_modules/dojo/has` module as a dependency to the intern config define function, changing the factory argument to a function, using `has('host-node')` and `has('host-browser')` conditionals and returning a dynamic configuration object - but that seems a bit too complex.

If anything, maybe more documentation should be added surrounding the default loader or `baseUrl`  values in all testing environments (both client and runner) since I have only read about them on stackoverflow or in issue threads.

jason0x43:
The PR allows the same `baseUrl` to work properly for `intern-client`, `intern-runner`, and `client.html`, assuming that intern is being run from the directory containing `node_modules` (the typical case).

- `basePath` is now a fully separate concept from `baseUrl`. Conceptually, `basePath` in the node environment is the same as the web server root when using client.html.
- `basePath` is a general Intern configuration parameter (vs. `baseUrl`, which is a loader parameter) that declares the root path from which files will be served
- `basePath` is only directly relevant to Node (runner, createProxy); it's ignored by the browser client. If not specified it defaults to process.cwd()
- If a relative `basePath` is specified in the intern config, it is relative to process.cwd()
- Relative `baseUrl`s are now relative to the directory containing `node_modules` when using `client.html`. This is different from typical dojo semantics, where a relative `baseUrl` is relative to the loading HTML doc, but it improves consistency between the node and browser environments. (In node environments a relative `baseUrl` is relative to process.cwd(), which is typically the directory containing `node_modules`.)

fixes #249

### DIFF
--- a/client.html
+++ b/client.html
@@ -5,32 +5,33 @@
 		<title>Browser test runner</title>
 		<script src="node_modules/dojo/dojo.js"></script>
 		<script>
-(function () {
-	var basePath = location.pathname.replace(/\/+[^\/]*$/, '/'),
-		internConfig = this.__internConfig = {
-			// For users running client.html directly, assume that the base path
-			// is two levels up (the parent of node_modules); this is most common.
-			// Users that need something special should provide an absolute module
-			// path for the config module (e.g. `/path/to/intern/config`) and
-			// then set `baseUrl` correctly in the `loader` configuration object
-			// of the config module
-			baseUrl: basePath.slice(-10) === '/__intern/' ? basePath.slice(0, -9) : (basePath + '../../'),
-			packages: [
-				{ name: 'intern', location: basePath }
-			],
-			map: {
-				intern: {
-					dojo: 'intern/node_modules/dojo',
-					chai: 'intern/node_modules/chai/chai'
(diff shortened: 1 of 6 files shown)


csnover:
This PR has bitrot and I’d like to get all the other great features from Intern 3 into peoples’ hands so I’m removing it from the milestone from now.

PR #310, opened 2014-11-21
Update client.html to increase error visibility

### LINKED ISSUE #230: Silent failure using `Date` in config file for IE <10
I'm using the geezer branch. I configured my Intern tests so each test would have a unique name usingn `Date.now()`. In my Intern test config file:

``` js
define({
    ...
    capabilities: {
        'selenium-version': '2.41.0',
        'name': 'mambocab  ' + Date.now()
    },
    ...
});
```

This works fine, even on the geezer branch, for ECMAScript 5 browsers. However, when `Date` isn't available in older browser, the test run just hangs without reporting any errors.

I have no problems with the failure itself, but it would have been easier to debug with some kind of error reporting. Is that feasible?

jason0x43:
An onerror handler is installed in client.html to catch errors thrown by Intern's loader. It simple and lightly styled.

fixes #230

### DIFF
--- a/client.html
+++ b/client.html
@@ -3,22 +3,35 @@
 	<head>
 		<meta charset="utf-8">
 		<title>Browser test runner</title>
+		<style>
+			.clientError {
+				padding: 0 20px;
+				border: solid 1px transparent;
+				font-family: sans-serif;
+			}
+			.clientError h2 {
+				background: #ffe0e0;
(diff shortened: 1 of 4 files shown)


jason0x43:
This will be taken care of in the [refactor-execution](https://github.com/theintern/intern/tree/refactor-execution) branch.

PR #312, opened 2014-11-26
fix duplicate suite reports with junit reporter

### LINKED ISSUE #275: JUnit reporter issues with Jenkins
This is a compilation of issues I have found with the JUnit reporter when testing with jenkins.

### Unit test results are duplicated.

【 code  】
- For a test suite with one test case, tested across three browsers, I get six test case results. Each of the test suite xml entries in report.xml corresponding to an actual unit test suite are preceded by one of:
 - `<testsuite name="firefox 31.0 on XP"...>`
 - `<testsuite name="chrome 37.0.2062.102 on XP"...>`
 - `<testsuite name="internet explorer 11 on WINDOWS"...>`
 - `<testsuite name="firefox 31.0 on XP"...><testsuite name="unit tests" ...><testsuite name="main"...>`
 - `<testsuite name="chrome 37.0.2062.102 on XP"...><testsuite name="unit tests" ...><testsuite name="main"...>`
 - `<testsuite name="internet explorer 11 on WINDOWS"...><testsuite name="unit tests" ...><testsuite name="main"...>`
【 end code 】

### No browser specifity reported

Jenkins seems to ignore nested testsuite elements and as such any browser specificity is lost in its junit test result output. Jenkins is grouping all testsuites into one package called (root). This can be split out per browser by using the classname attribute as specified here: http://nelsonwells.net/2012/09/how-jenkins-ci-parses-and-displays-junit-output/

I will potentially make a pull request to fix this but would like to know what the current JUnit reporter output is being tested against so that regressions aren't introduced.

nicknisi:
When the junit reporter is run with the runner, the reports are duplicated. It appears that this is caused by the `/suite/end` topic being published by both `Suite` and `createProxy` (in `publishInSequence`). This appears to only be an issue with the junit reporter, as it appears to count on `/suite/end` to only be called once per suite. This fixes the duplication reported in fixes #275. This change should not affect the junit reporter when executed with client (unit test still passes).

### DIFF
--- a/lib/reporters/junit.js
+++ b/lib/reporters/junit.js
@@ -138,7 +138,11 @@ define([
 				return;
 			}
 
-			rootNode.childNodes.push(createSuiteNode(suite, sessions[suite.sessionId || '']));
+			var runId = '_' + suite.sessionId + '-run';
+			if (!this[runId]) {
+				this[runId] = true;
+				rootNode.childNodes.push(createSuiteNode(suite, sessions[suite.sessionId || '']));
+			}
 		},
 
 		start: function () {


csnover:
This is fixed in master

PR #330, opened 2015-01-09
Add the option to specify the report location for junit reporter

Zensavona:
Optionally specify a `junitReportFile` in your configuration file to output to a specific location, or if unset, default to `report.xml` as before.

### DIFF
--- a/lib/reporters/junit.js
+++ b/lib/reporters/junit.js
@@ -6,7 +6,7 @@ define([
 	'intern',
 	'dojo/node!fs',
 	'../util'
-], function (fs, util) {
+], function (intern, fs, util) {
 	var sessions = {
 		'': 'Node.js'
 	};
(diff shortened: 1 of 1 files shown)


sholladay:
I would personally prefer a more standardized API for report locations, de-coupled from which exact reporter is being used. This could get very annoying when implementing a new reporter or switching between them as a user of Intern.

csnover:
Sorry, I totally missed this since I only double-check the issues list right now.

We are enhancing the reporting system to be fully configurable for the next release, so this patch can be used by people as a stop-gap but isn’t likely to make it into any release. You can see progress of our reporting system enhancements in [this file in the refactor-execution branch](https://github.com/theintern/intern/blob/refactor-execution/lib/ReporterManager.js).

Zensavona:
All good, I just patched it for my own use (I needed it immediately) so I figured may as well share if you guys want it also.

It would be nice to have a more regular way of doing this for all reporters though.

Cheers

csnover:
Good news! Once the [refactor-execution](https://github.com/theintern/intern/tree/refactor-execution) branch is finished and landed for Intern 3, all reporters in Intern will become configurable. As such I am going to close this now in order to reduce the number of outstanding PRs. Thank you for your patches!

PR #343, opened 2015-01-30
added the possibility to pass NODE_ENV environment variable via grunt config

leohuber:
Allows to pass a variable 'nodeEnv' (NODE_ENV) environment variable from the grunt config and to run intern tests with different config variables for different environments using https://github.com/lorenwest/node-config.


csnover:
Thanks for the patch! We’re currently focused on getting the [refactor-execution](https://github.com/theintern/intern/tree/refactor-execution) branch finished and landed and then will be evaluating all open PRs.

Issue #329, opened 2015-01-08
Feature request - pass in a custom value for the istanbul coverageVariable setting


plumlee:
Love writing and running test with intern - thanks for such a great tool.

Would you consider adding a configuration option to allow setting the Istanbul coverageVariable to a custom value?

In my case, I  am using intern+browserify+browserify-istanbul to do my testing. I'm currently passing the internal variable of `__internCoverage` in order to prevent multiple instrumentation of files, but I'd prefer to not depend on a private setting to do so. It doesn't appear that this variable is exposed in any way outside of direct reading of source code (I used https://github.com/theintern/intern/wiki/Using-Custom-Arguments-and-Configuration-Properties to read the config as a quick test).

If you would prefer not to allow setting of this value, then could it be exposed in the configuration somehow?

csnover:
We would be happy to make the coverage variable a configurable option. If you would like, please submit a PR containing the changes, otherwise we will look at scheduling this work for a future release. Note that we are working on some [major refactoring](https://github.com/theintern/intern/tree/refactor-execution) right now so it will be a little while before a PR would be accepted, and it would probably need to be modified a bit once the refactor-execution branch is merged into master. As long as the PR is in place though that makes it a lot easier for us to move forward with finishing the feature.

plumlee:
Will work on this and get a PR together - thanks.

plumlee:
Running into a bit of an issue that I will eventually resolve, but thought I'd ask here: 

https://github.com/theintern/intern/blob/master/runner.js#L282

Not sure in what scope `__internCoverage` is defined here - I've tried a few different things but it's not clear to me. Seems that it's being inherited from the Tunnel but not sure. I'm trying to do something like 

```
if (scope && typeof scope[config.coverageVariable] !== 'undefined') ......
```

but I'm not getting scope correct.

vladikoff:
@plumlee do you mean `global.` ? http://nodejs.org/api/globals.html#globals_global

plumlee:
I've tried `global` but it's not defined there (the property of global, I mean - global[config.coverageVariable]). At least when running it like this: `~/git/intern_MASTER/bin/intern-runner.js config=test/config/intern-unit-browser` where intern_MASTER is my local forked copy. Tests run, but I do't see the output. I can go ahead push what I have so far to my fork if that's helpful to look at.

plumlee:
[...]

PR #344, opened 2015-01-30
Allow setting the code coverage variable name

plumlee:
Implements #329.


csnover:
Thanks for the patch! We’re currently focused on getting the [refactor-execution](https://github.com/theintern/intern/tree/refactor-execution) branch finished and landed and then will be evaluating all open pulls.

PR #351, opened 2015-02-15
Preserve internal mappings for modules, fixes #348

### LINKED ISSUE #348: Defining a loader map property of ‘*’ prevents intern runner from loading dojo node module
A user-defined loader `map` property with a `*` property means you have to redefine the `intern/dojo` path yourself.  If not, you get the following when trying to run a test:
```
$ ./node_modules/.bin/intern-runner config=test/config/intern-functional
FATAL ERROR
Error: Failed to load module intern/dojo/node from /Users/path/to/repo/node_modules/intern/dojo/node.js (parent: intern/dojo/node!38!*)
  at <node_modules/intern/node_modules/dojo/dojo.js:757:12>
  at <fs.js:208:20>
  at Object.oncomplete  <fs.js:108:15>
```

I’m using a base configuration for intern like so:

```
    loader: {
        map: {
            '*': {
                'chai'       : '../../../node_modules/chai/chai',
                'sinon'      : '../../../node_modules/sinon/pkg/sinon',
                'sinon-chai' : '../../../node_modules/sinon-chai/lib/sinon-chai'
            }
        }
    },
```

Then I override particular settings for each kind of test I run (unit, functional) by extending this config. When running a functional test like using the command and the test example below, I get the error shown above.

This line - https://github.com/theintern/intern/blob/2.2.2/runner.js#L83 - doesn’t preserve the default mapping that intern expects, as defined earlier in the same file: https://github.com/theintern/intern/blob/2.2.2/runner.js#L13

Seems like runner.js should load with that base config, then reload itself and merge my config with the defaults for  `map` to preserve the needed intern paths.

Feels like it could get really complicated, trying to figure out what mappings the user might have overwritten, and what mappings intern should keep. I’m wondering if the simplest fix is to add a line of documentation here: https://github.com/theintern/intern/wiki/Configuring-Intern#loader
that states you must map the `intern/dojo` path if you defined a mapping for ‘*’, or maybe that you must redefine any intern paths that you want to reuse, such as for chai or dojo loaded by intern. 

 In the end, it’s not really a bug so much as a non-obvious possible side effect of defining your own loader configuration. I don’t think a user should have to know how intern defined its own paths internally and when those have to be redefined, but it’s not out of the question to expect a user to understand the implications if it’s noted in the documentation.

I’m happy to throw up a test repo if that would make this clearer. Also, if you’ll let me know what solution you’d prefer, I’ll do a pull request.

plumlee:
Fixes #348. Believe this is sufficient to avoid the problem.

### DIFF
--- a/runner.js
+++ b/runner.js
@@ -1,4 +1,17 @@
 /*jshint node:true, es3:false */
+
+// need to preserve our own mappings in case use specifies them as well
+var amdMappings = {
+	intern: {
+		dojo: 'intern/node_modules/dojo',
+		chai: 'intern/node_modules/chai/chai',
+		diff: 'intern/node_modules/diff/diff'
+	},
+	'*': {
+		'intern/dojo': 'intern/node_modules/dojo'
+	}
+};
+
 if (typeof process !== 'undefined' && typeof define === 'undefined') {
 	(function () {
 		// this.require must be exposed explicitly in order to allow the loader to be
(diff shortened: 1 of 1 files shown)


csnover:
Thanks for the patch! We’re currently focused on getting the [refactor-execution](https://github.com/theintern/intern/tree/refactor-execution) branch finished and landed and then will be evaluating all open pulls.

PR #354, opened 2015-02-24
fix typo in the 'Who is Intern best for?' section 'other' should be 'oth...

modcoms:
...erwise'


csnover:
I think this change really is a wise one.

http://www.sadtrombone.com/

Thanks for your help!

PR #378, opened 2015-04-10
Update intern config example

kospiotr:
Make exclude instrumentation also working on Windows


csnover:
It’s no longer necessary to use backslashes in Windows, so this config change is not necessary! :+1:

PR #383, opened 2015-04-20
Qunit Interface with tests

apsdehal:
- Add QUnit lib interface for intern.
- Contains tests for the interface.


csnover:
CLA is on file.

PR #387, opened 2015-05-05
Update Chai to latest version

andyearnshaw:
The current version of Chai included with Intern is considerably out of date now, with features available in the docs that are unavailable in Intern's version.  This has caught me out a few times, so I'm hoping it can be updated smoothly.

I've tested updating my local Intern's Chai dependency to 2.3.0 and haven't encountered any problems, but I'm unsure how to run the tests on intern itself.


csnover:
Thanks!

PR #395, opened 2015-05-18
Fix for unit tests freezing in browser

cezarykluczynski:
Since ac46e32 unit tests run in browser freeze after they are done executing, and no other page is loaded. This commit fixes it for me.

### DIFF
--- a/lib/ClientSuite.js
+++ b/lib/ClientSuite.js
@@ -56,6 +56,8 @@ define([
 							suite.tests.forEach(function (test) {
 								self.tests.push(test);
 							});
+
+							receiveEvent('runEnd');
 						}
 						else {
 							forward();


csnover:
runEnd should not be dispatched this way, it should be sent by the client after the test run has completed. I don’t know how the final suiteEnd would be sent but then runEnd is missing. Do you have a reproduction case?

cezarykluczynski:
I've been using Intern here: https://github.com/cezarykluczynski/qunit-desktop-notifications Now there's Intern 2.2.2 in package.json, cause I downgraded, but switching back to master, installing (npm install && bower install) and running grunt test, assuming selenium-standalone is running locally, will cause the freeze.

csnover:
Thanks, I’ve reproduced with your repo and am investigating!

csnover:
@cezarykluczynski I’m not able to reproduce this after updating to the latest release of Chrome. Could you please make sure your Chrome installation is up-to-date (43.0.2357.65) and try again and let me know what happens?

cezarykluczynski:
With the latest master and Chrome 43, this issue is gone (it was indeed present in Chrome 42). Thanks!

PR #396, opened 2015-05-21
Match backslash as separator in excludeInstrumentation.

tonygambone:
This makes the default value function correctly on Windows.


csnover:
See #378.

PR #402, opened 2015-05-30
Auto-glob in Node.js

### LINKED ISSUE #359: Auto-glob in Node.js
When running in Node.js, we have filesystem access, so should allow glob patterns in `suites` and `functionalSuites` for users that don’t ever run tests directly in the browser.

cezarykluczynski:
I'm not really sure if that's the right approach, but please consider.

Probably fixes #359.

### DIFF
--- a/lib/executors/PreExecutor.js
+++ b/lib/executors/PreExecutor.js
@@ -5,11 +5,12 @@ define([
 	'dojo/Promise',
 	'dojo/has!host-browser?dojo/request',
 	'dojo/has!host-node?dojo/node!path',
+	'dojo/has!host-node?dojo/node!glob',
 	'../../main',
 	'../parseArgs',
 	'../util',
 	'require'
-], function (aspect, has, lang, Promise, request, pathUtil, main, parseArgs, util, require) {
+], function (aspect, has, lang, Promise, request, pathUtil, globUtil, main, parseArgs, util, require) {
 	/**
 	 * For testing sessions running through the Intern proxy, tells the remote test system that an error occured when
 	 * attempting to set up this environment.
@@ -219,6 +220,25 @@ define([
 						config[name] = value === '' ? [] : [ value ];
 					}
 				});
+
+				if (has('host-node')) {
+					var files, splice;
+					[ 'suites', 'functionalSuites' ].forEach(function (name) {
+						config[name].forEach(function (pattern, index) {
+							files = globUtil.sync(pattern);
+
+							// Original value is replaced by glob findings only if something was found.
+							// This allow mixing glob patterns with regular paths.
+							if (files.length > 0) {
+								splice = [index, 1];
+								files.forEach(function (filename) {
+									splice.push(filename);
+								});
+								Array.prototype.splice.apply(config[name], splice);
+							}
+						});
+					});
+				}
 
 				return config;
 			});
--- a/lib/executors/PreExecutor.js
+++ b/lib/executors/PreExecutor.js
@@ -221,24 +221,32 @@ define([
 					}
 				});
 
-				if (has('host-node')) {
-					var files, splice;
-					[ 'suites', 'functionalSuites' ].forEach(function (name) {
-						config[name].forEach(function (pattern, index) {
-							files = globUtil.sync(pattern);
-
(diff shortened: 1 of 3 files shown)


lib/executors/PreExecutor.js
```diff
@@ -220,6 +221,25 @@ define([
...
+					var files, splice;
+					[ 'suites', 'functionalSuites' ].forEach(function (name) {
+						config[name].forEach(function (pattern, index) {
+							files = globUtil.sync(pattern);
+
+							// Original value is replaced by glob findings only if something was found.
+							// This allow mixing glob patterns with regular paths.
+							if (files.length > 0) {
+								splice = [index, 1];
+								files.forEach(function (filename) {
```
csnover: `splice = [ index, 1 ].concat(files);`

lib/executors/PreExecutor.js
```diff
@@ -220,6 +221,25 @@ define([
 					}
 				});
 
+				if (has('host-node')) {
+					var files, splice;
```
csnover: One `var` per var, please (code style preference). It would also be better to define these further down, inside the anonymous function where they are used.

lib/executors/PreExecutor.js
```diff
@@ -220,6 +221,25 @@ define([
...
+							files = globUtil.sync(pattern);
+
+							// Original value is replaced by glob findings only if something was found.
+							// This allow mixing glob patterns with regular paths.
+							if (files.length > 0) {
+								splice = [index, 1];
+								files.forEach(function (filename) {
+									splice.push(filename);
+								});
+								Array.prototype.splice.apply(config[name], splice);
```
csnover: You’ll either need to use `config[name].slice(0).forEach` above, or wait to splice in changes ’till after you’ve finished running through the list. It’s easiest to explain why by example:

```js
var array = [ 'a', 'b', 'c' ];
array.forEach(function (item, index) {
  console.log(item);
  if (item === 'b') {
    array.splice(index, 1, 'd', 'e', 'f', 'g');
  }
});
```

logs 'a' 'b' 'e' because the 3rd item in the array ('c') becomes 'e', but we wanted to look at 'c'.

csnover:
Thank you very much for sending this! There are only a couple of minor problems that should be easy to fix (in inline comments), plus a couple final considerations that don’t necessarily need to be addressed in this PR (but it would be great if they were!):

1. If glob patterns are encountered in the browser for `suites`, an error should be thrown to let users know that globbing only works in Node.js.

2. We should be able to expand the globs in Node.js and send them to the client in the test runner. To avoid problems with URL length, the best way I can think to do this would be to extend `lib/Proxy.js` to listen for a special path (like `/__intern/config.js`) that will send something like this:

   ```
define([ 'originalConfigId' ], function (config) {
  config.suites = [ /* list of globbed suites */ ];
  return config;
});
```

   Then, change `lib/ClientSuite.js` to tell the browser client to load this special config instead.

cezarykluczynski:
@csnover Thanks for reviewing! I hope to fix the raised issues in a two or three days.

Thanks for pointing out the splice thing. I don't know what i was thinking, modyfing array in the middle of iteration...

cezarykluczynski:
@csnover I think that's it. I'm afraid I won't be able to do sending expanded of globs to browser. There just too much code to go through. I also found and fixed one error in tests. I've tried to write some unit tests for PreExecutor, but it was hard to separate PreExecutor from the rest of code (util.getModule in getConfig was what got me).

#405 is another thing I found.

csnover:
@cezarykluczynski OK, if you want guidance to take this to completion you can talk to some folks on IRC otherwise I can try to take over (the second one might take longer). Let me know. Thanks!

cezarykluczynski:
@csnover I'll pass. Please take over.

jason0x43:
Resolved by 86193c293977ad70c74b0c0877fe2028c8d6a1c2

vladikoff:
@jason0x43 Awesome!! Do we need to update the docs as well (which is basically copy pasting your commit message)?

jason0x43:
I added some basic usage (not implementation) information to the `gh-pages-next` branch. Take a look and let me know if you think it needs more.

PR #416, opened 2015-06-13
Fix #415 by adding QUnit.extend method

### LINKED ISSUE #415: QUnit Interface: Implement QUnit.extend
This is a must need, I will soon put up a patch for this.

apsdehal:
Add QUnit.extend method for QUnit interface along with tests.

### DIFF
--- a/lib/interfaces/qunit.js
+++ b/lib/interfaces/qunit.js
@@ -154,13 +154,10 @@ define([
 		extend: function (a, b, undefOnly) {
 			for (var prop in b) {
 				if (b.hasOwnProperty(prop)) {
-					// Avoid "Member not found" error in IE8 caused by messing with window.constructor
-					if (!(prop === "constructor" && a === window)) {
-						if (b[ prop ] === undefined) {
-							delete a[ prop ];
-						} else if (!(undefOnly && typeof a[ prop ] !== "undefined")) {
-							a[ prop ] = b[ prop ];
-						}
+					if (b[prop] === undefined) {
+						delete a[prop];
+					} else if (!(undefOnly && typeof a[prop] !== "undefined")) {
+						a[prop] = b[prop];
 					}
 				}
 			}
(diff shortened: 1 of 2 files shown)


lib/interfaces/qunit.js
```diff
@@ -151,6 +151,23 @@ define([
 			testTimeout: Infinity
 		},
 
+		extend: function (a, b, undefOnly) {
+			for (var prop in b) {
+				if (b.hasOwnProperty(prop)) {
+					// Avoid "Member not found" error in IE8 caused by messing with window.constructor
```
csnover: ie8 is not supported so this is not necessary

lib/interfaces/qunit.js
```diff
@@ -151,6 +151,23 @@ define([
...
+					if (!(prop === "constructor" && a === window)) {
+						if (b[ prop ] === undefined) {
```
csnover: no spaces in array accessors please :)

tests/unit/lib/interfaces/qunit.js
```diff
@@ -351,6 +351,18 @@ define([
 			}
 		},
 
+		'extend': {
+			'should have a working expect': function () {
```
csnover: this test doesn’t test the various `undefined` handling branches so is not complete.

tests/unit/lib/interfaces/qunit.js
```diff
@@ -351,6 +351,18 @@ define([
...
+				QUnit.extend( QUnit.assert, {
```
csnover: this test is mutating objects that are not reset! (QUnit.assert)

csnover: use a new object scoped to the test to test this utility function, there is no reason to run it against anything constructed outside of the test function.

apsdehal: Fixed!

csnover:
Thanks for the patch!

PR #428, opened 2015-06-26
Fix Description of Test.timeout

fcarstens:
As per discussion on IRC, timeout is the time that the suite is allowed to take to execute.


csnover:
Thanks for the patch!

PR #430, opened 2015-06-29
Report per-suite skips in the Runner reporter.

### LINKED ISSUE #421: Platform part of the "tests failed" report is missing how many tests were skipped
Description: Platform part of the "tests failed" report is missing how many tests were skipped

### Steps to reproduce: Platform "tests failed" report is missing how many tests were skipped

1. run some tests with the `grep` option, I used `npm run test-functional grep="sync reset password, verify same browswer"` 
2. Got output: 
```
> fxa-content-server@0.39.0 test-functional /Users/vladikoff/mozilla/fxa-local-dev/fxa-content-server
> intern-runner config=tests/intern_functional "grep=sync reset password, verify same browswer"

Creating Firefox profile...
Listening on 0.0.0.0:9090
Tunnel started
‣ Created session firefox on any platform (cf6916d0-9e34-47f6-b3cf-85688b7abb77)
Waiting for email at: http://127.0.0.1:9001/mail/signin0.31952479644678533
Waiting for email at: http://127.0.0.1:9001/mail/signin0.2677939438726753
Waiting for email at: http://127.0.0.1:9001/mail/signin0.4526257694233209
Waiting for email at: http://127.0.0.1:9001/mail/signin0.8374885404482484
✓ firefox on any platform - Firefox Desktop Sync reset password - sync reset password, verify same browswer (4.933s)
No unit test coverage for firefox on any platform
firefox on any platform: 0/264 tests failed


TOTAL: tested 1 platforms, 0/264 tests failed (263 skipped)
```

### Expected result:
The reporter before the `TOTAL` should include skipped tests or just say how many tests ran:
`firefox on any platform: 0/1 tests failed` or  `firefox on any platform: 0/264 tests failed (263 skipped)`
### Actual result:
Says `firefox on any platform: 0/264 tests failed` and does not include skipped tests.

Intern version: Intern 3 at Git Hash : 42aebd9beb942a11e2c6e6d7687c70a1c22b9bf7

Any additional information: None

sholladay:
Closes #421.

### DIFF
--- a/lib/reporters/Runner.js
+++ b/lib/reporters/Runner.js
@@ -160,11 +160,15 @@ define([
 						.write('\n');
 				}
 
-				var numFailedTests = suite.numFailedTests;
+				var name            = suite.name;
+				var numFailedTests  = suite.numFailedTests;
+				var numTests        = suite.numTests;
+				var numSkippedTests = suite.numSkippedTests;
+				var message         = '%s: %d/%d tests failed (%d skipped)';
 
 				this.charm
 					.foreground(numFailedTests > 0 ? LIGHT_RED : LIGHT_GREEN)
-					.write(suite.name + ': ' + suite.numFailedTests + '/' + suite.numTests + ' tests failed')
+					.write(nodeUtil.format(message, name, numFailedTests, numTests, numSkippedTests))
 					.display('reset')
 					.write('\n\n');
 			}


sholladay:
Working on this, the value that is assigned to suite.name seems super weird. It is not the name of the test suite whatsoever. But I didn't look into it at all - maybe the variable called `suite` isn't really what it says it is and its just badly named?

From IRC:
````
13:45:15	<sholladay>	it's strange that suite.name is "firefox on any platform"
13:45:57	<vladikoff>	sholladay, a bit, i guess if you have 2 browsers going then it makes more sense
13:46:15	<vladikoff>	sholladay, also "any platform" should say Linux but there is probably some configuration required
13:47:27	<sholladay>	Yeah, I would expect it to get computed and live at something like suite.environment
13:49:15	<sholladay>	or maybe not on the suite, per se, but the suite has a name defined in the test module that doesn't get put there, that's the real weirdness
````

csnover:
Re: `suite.name`, the test runner creates one root suite corresponding to each environment that you are running against. The names of these suites are generated before the server says what the actual environment is, so if you don’t provide explicit version information in your `environments` config it will end up being “any”, since that’s what was specified. It would be nice to mutate this once the information becomes available but changing the name changes the id of all sub-suites/tests, which could lead to some bad stuff happening, so if/until someone has time to figure out what might break it’s just a little mysterious if you don’t give explicit values to your environment.

csnover:
Thank you for the patch! I made a slight adjustment to match the way skipped tests are reported at the end of a run, which is to say the skipped part only appears when there are actually any skipped tests at all.

sholladay:
Not sure how I overlooked that. I like how it ended up. :)

Issue #424, opened 2015-06-23
google fonts coming in 404 in client.html


httpete:
Description:

We have a secure environment, and I can't load ANY external resources. Can we disable somehow or maybe bundle in the google fonts to make client.html look so pretty? It really slows down the client.html test page, since it waits for the google font to 404.

Intern version: 3.0 RC1

Any additional information:

csnover:
We’ve discussed this internally and won’t be adding font bundling since the edge case nature of this problem doesn’t really justify the extra space that these fonts would take for *every* Intern user. You could write and use a [custom](https://theintern.github.io/intern/#custom-reporters) version of the HTML reporter that includes the fonts if this is a significant problem for your organisation.

httpete:
Then don't add the font from google at all please, it's not worth it. Secure environments are the new norm - people will not be able to import things from the interwebs. It makes the whole test client.html hang..

httpete:
@csnover @dylans if there is any way I could gently bump on this - it is going to make adoption harder in my org. The html browser reporter is so nice, and just because we are adding fonts from google I have to jump through hoops. It looks sweet without the fonts, it is so subtle.

dylans:
I created a fairly simple PR for this. See my notes and question. I agree that this makes Intern simpler for people to use inside the firewall for testing.

httpete:
Thank you. This is a good heads up - we are now in a COMPLETE secure VED virtual dev env. I think this will happen to more organizations - the hacking out there is EPIC and is getting worse. We are now in a locked - no interwebs access environment. Makes life wicked hard.

dylans:
It's been reported that this causes an issue when trying to run tests on a simulated slow connection to more accurately reflect mobile testing on a slow network.

httpete:
The sizzle of the font isn't worth this IMHO

httpete:
I am preparing to showcase this, and I have to wait till this times out until I can show the sexy ui inside my protected corporate environment :-(

jason0x43:
The google fonts were removed in 0fb9352b.

PR #433, opened 2015-07-08
refs #424, split third party fonts into configuration option

dylans:
A fairly simple solution for #424... given that Intern is used inside of corporate intranets, it seems like a reasonable configuration option to provide.

My check is naive (and assumes that window.location exists), but I figured I'd submit this first, and see if the better option is to include the user's config, and read a setting from there.

### DIFF
--- a/lib/reporters/Html.js
+++ b/lib/reporters/Html.js
@@ -122,7 +122,10 @@ define([
 
 			var link = document.createElement('link');
 			link.rel = 'stylesheet';
-			link.href = require.toUrl('./html/html.css');
+			// TODO: Could this be read from the config instead of URL?
+			var href = (/intranet/.test(location.search)) ?
+				'./html/html-intranet.css' : './html/html.css';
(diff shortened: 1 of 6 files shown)


csnover:
Reporters should be configured through their passed options hash as described in [custom reporters](http://theintern.github.io/intern/#custom-reporters), not in the way this patch works, so I am closing this PR. I also don’t actually want to accept any change related to that ticket until I hear from more than one reliable source that it’s a problem, which is why I marked it as wont-fix to begin with. (If a 404 is actually returned for a font resource there is no FOUT so the report makes no sense.)

httpete:
It absolutely does FOUT. it is white for 1 minute and then shows the ui.

jason0x43:
The Google font will be removed, probably in the next point release, and definitely in the next major release. Visual consistency across platforms is nice, but not _that_ important.

httpete:
Jason thank you it looks great and now will load snappily

PR #434, opened 2015-07-08
Add '=' sign to -Dwebdriver command line argument example

cwaldren:
Couldn't get it to work at first with Selenium throwing "poorly formatted Java property setting". Realized the equals sign was needed when writing -Dwebdriver.chrome.driver=chromedriver.exe


csnover:
Thanks!

PR #451, opened 2015-08-01
Handle WebDriver DOM Output Better

### LINKED ISSUE #435: Make WebDriver reporter more resilient against DOM-nuking tests when writeHtml is true
I am getting the following error when running certain tests under runner:

```
Reporter error!
TypeError: Unable to get property 'parentNode' of undefined or null reference
at suiteEnd  <__intern/lib/reporters/WebDriver.js:51:5>
at Anonymous function  <__intern/lib/ReporterManager.js:229:7>
at emit  <__intern/lib/ReporterManager.js:211:4>
at report  <__intern/lib/Suite.js:199:6>
at end  <__intern/lib/Suite.js:193:5>
at Anonymous function  <__intern/node_modules/dojo/Promise.js:156:33>
at runCallbacks  <__intern/node_modules/dojo/Promise.js:19:13>
at Anonymous function  <__intern/node_modules/dojo/Promise.js:103:21>
at run  <__intern/node_modules/dojo/Promise.js:51:33>reset
``` I suspect this is because I am modifying the DOM myself and the WebDriver reporting is getting a bit lost and is accessing the elements in an "unsafe" way.

In particular, wouldn't [this line](https://github.com/theintern/intern/blob/master/lib/reporters/WebDriver.js#L51):

```js
this.suiteNode = this.suiteNode.parentNode.parentNode || document.body;
``` Be more safely written as:

```js
this.suiteNode = this.suiteNode && this.suiteNode.parentNode && this.suiteNode.parentNode.parentNode || document.body;
```

kitsonk:
- WebDriver outputs to a `<div>` which, if isn't attached to `document.body` when a suite finishes, will get reattached.  Prevents interference from other changes to the DOM by tests.
- Add unit test coverage for WebDriver reporter.

Fixes #435


kitsonk:
While the changes to to the WebDriver reporter are still mergable, the unit tests for the WebDriver reporter (which don't exist in master) are a bit hopelessly beyond repair.  The original behaviour prior to the 3.0.4 refactoring is that when `waitForReporter` was `false`, the messages would be sent synchronously to request.  Now, the requests are sent asynchronously no matter what, but the only change is that when waitForReporter is false, it does not return the promise.  When `waitForReporter` is `true`, a promise is returned, which allows the runner to wait until the promise is resolved before continuing.  The unit tests use a mock for request, but the problem now is that because the request is not resolved in the same turn, the request data is never available for introspection when the test continues, but of course because there is not Promise returned, there is no reliable way to know exactly when the request has been sent.

Of course, I could just hack the unit tests to wait a bit to introspect the request mock, but that doesn't seem to be the best way of handling things.

kitsonk:
Actually, correction, it was a different change in behaviour... what gets posted to request is the data payload wrapped into an array.  Previous behaviour was only sending the request data payload as a JSON object serialized.  Updating the unit tests to reflect the change in behaviour.

kitsonk:
Ok, double back on that... when more then one event is issued during the same turn, the first one is resolved synchronously, the second one is resolved in the next turn and there is no handle to that.  Working around the issue for the time being, but clearly there is no way of "knowing" when the messages get dispatched at the moment from the public WebDriver API.

kitsonk:
I can't figure out why this is failing... The self tests don't actually report any errors, but it marks all the test cases in SauceLabs as failures.  It could be because I actually do throw errors in the WebDriver reporter to verify it does represent those correctly, but I still can't figure out why it is still exiting badly without reporting the error console side.

dylans:
We should rebase and review, and decide what we're going to do with this one.

kitsonk:
It is current with master already...

jason0x43:
Merged in e4d461b

Issue #455, opened 2015-08-04
Proxy.js fails to serve path ending in '/' on Windows


gitgrimbo:
The `Proxy.js` code only checks for `'/'` (https://github.com/theintern/intern/blob/3.0.0/lib/Proxy.js#L98):

`if (wholePath.charAt(wholePath.length - 1) === '/') {`

But on Windows, this last character would be `'\'`.

#### For example

Requested URL = `"http://somehost:9000/webapp/"`
wholePath = `"d:\dev\git_repos\myproject\webapp\"`
url = `"/webapp/"`
file = `"webapp/"`

Results in the error:

````
Error: EISDIR, read
  at Error  <native>
npm ERR! Test failed.  See above for more details.
````

#### Fix

Use `path.sep`:

`if (wholePath.charAt(wholePath.length - 1) === path.sep) {`

PR #456, opened 2015-08-04
refs #455, use platform path.sep in Proxy.js

gitgrimbo:
#455 

The `Proxy.js` code only checks for '/':

`if (wholePath.charAt(wholePath.length - 1) === '/') {`

But on Windows, this last character would be '\'.

So use path.sep in comparison.


csnover:
4c480525515924422610c3378478664c4d21acce. Thanks for the patch!!

PR #484, opened 2015-09-04
Add message buffering

jason0x43:
This PR changes how lifecycle messages are sent from the client to Intern. Currently, each message is sent in an individual request. Doing so can result in many requests being active at a time, which can starve other parts of the testing system that involve requests (like tests). With this PR, only one message request is ever active at a time. While a request is in progress, new messages are buffered. As soon as the current request completes, a new one is started to send all the messages buffered up to that point.

### DIFF
--- a/lib/Proxy.js
+++ b/lib/Proxy.js
@@ -59,10 +59,17 @@ define([
 
 				var self = this;
 				request.on('end', function () {
-					var message = JSON.parse(data);
-					var runnerReporterPromise = self._publishInSequence(message);
+					var messages = JSON.parse(data).map(function (messageString) {
+						return JSON.parse(messageString);
+					});
 
-					var shouldWait = util.getShouldWait(self.waitForRunner, message);
+					var runnerReporterPromise = Promise.all(messages.map(function (message) {
+						return self._publishInSequence(message);
+					}));
+
+					var shouldWait = messages.some(function (message) {
+						return util.getShouldWait(self.waitForRunner, message);
+					});
 
 					if (shouldWait) {
 						runnerReporterPromise.then(function () {
--- a/lib/reporters/WebDriver.js
+++ b/lib/reporters/WebDriver.js
@@ -24,9 +24,7 @@ define([
 			this.suiteNode = document.body;
 		}
 
-		if (this.waitForRunner) {
-			this.enqueue = util.createQueue(4);
-		}
+		this._messageBuffer = [];
 	}
 
 	WebDriver.prototype = {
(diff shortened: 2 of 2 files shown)


bryanforbes:
I've tested this with dojo-core and all tests pass (except those that we know should be failing).

csnover:
When you say starve, do you just mean that other things are slow because they are temporarily blocked or something else? The test system should either fill the browser’s outgoing connection limit but in a way where the requests are short so other things can be safely interleaved (no waitForRunner), or it should limit to 4 outstanding requests (waitForRunner). Are you seeing some other behaviour? I am OK with the buffering approach except am concerned this patch will just introduce result latency since I don’t know what it would fix otherwise.

jason0x43:
By starve I do mean that things are temporarily blocked. What I've observed is that Intern can queue up many (10+) requests at a time, because even though they're small, they still take a while to complete. The flood of test messages through much of the testing process can cause resources loaded by the tests to be delayed. (This is complicated even more in IE which doesn't always appear to process queued requests in order, and will occasionally abort queued requests for no discernible reason.) For example, in the dojo2 tests, a file loaded in one of the tests typically takes more than 5s in IE, which causes a test timeout. Most of that time is spent waiting for the request to be processed.

Buffering the results prevents the above issues because it minimizes the connection resources being used by the test system (leaving more for the tests). It has a couple other advantages, too: it handles network latency much better than sending a flood of small requests, and it can simplify how Intern processes incoming messages because they're guaranteed to always be in order (this wasn't taken advantage of in the PR). Buffering can increase result latency somewhat, although I don't think it's that significant -- over time, all messages arrive back at Intern as fast or faster than they did when sent individually, so any particular message shouldn't be significantly delayed.

csnover:
OK, the delay seems like a reasonable trade-off. This should probably also remove the old queue code too, right, since there’s no reason to have the queue any more (there’s always only one request out to XHR no matter what)?

jason0x43:
The PR does remove the queue code from the WebDriver reporter.

csnover:
Ah, so it does. Well there is a reason why I say I am burnt out :)

csnover:
As a side note, we’re going to have to start dealing with reducing the memory utilisation of messages transmitted to the server Real Soon.

jason0x43:
Like, the proxy is consuming large amounts of memory?

csnover:
Well,

* Messages are serialised at the moment they are received by the reporter,
* Suite messages contain a serialisation of all their child tests as well,
* Those strings are persisted in an array on the client, then that array gets serialised a second time before being passed on for transmission.

So it’s quite a bit of memory (and bandwidth!) being used for these messages relative to how much more efficient it *could* be.

jason0x43:
Oh, all *that* memory usage.

Issue #500, opened 2015-09-22
fixdeps.js symlink error Win7 npm 3


gwicksted:
Possibly related to #429.

Node `4.1.0` with npm `3.3.3` on `Windows 7 x64`

```sh
npm ERR! Windows_NT 6.1.7601
npm ERR! argv "C:\\Program Files (x86)\\nodejs\\node.exe" "C:\\Users\\myusername\\AppData\\Roaming\\npm\\node_modules\\
npm\\bin\\npm-cli.js" "install" "intern" "--save-dev"
npm ERR! node v4.1.0
npm ERR! npm  v3.3.3
npm ERR! code ELIFECYCLE

npm ERR! intern@3.0.3 install: `node support/fixdeps.js`
npm ERR! Exit status 1
```

Further up, more details revealed:

```sh
fs.js:902
  return binding.symlink(preprocessSymlinkDestination(destination, type, path),
                 ^

Error: EPERM: operation not permitted, symlink 'D:\dev\myprojectfolder\node_modules\dojo' -> 'D:\dev\myprojectfolder\node_modules\intern\node_modules\dojo'
    at Error (native)
    at Object.fs.symlinkSync (fs.js:902:18)
    at D:\dev\myprojectfolder\node_modules\intern\support\fixdeps.js:33:6
    at Array.forEach (native)
    at Object.<anonymous> (D:\dev\myprojectfolder\node_modules\intern\support\fixdeps.js:14:28)
    at Module._compile (module.js:434:26)
    at Object.Module._extensions..js (module.js:452:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Function.Module.runMain (module.js:475:10)
```

gwicksted:
Paths need normalizing before performing a string comparison for equality.  The two offenders:

```js
var expected = path.join(__dirname, '..', 'node_modules');
```

```js
	var actualPath = path.dirname(require.resolve(path.join(dependency, 'package.json')));
```

a `path.normalize` ensures that "C:\" and "c:\" will be the same case before performing string comparisons.  I believe it's __dirname that returns the inconsistent case.

csnover:
Copying over from the closed PR #501…

The issue here is:

1. npm 3 doesn’t ever install dependencies into the intern/node_modules` directory so everyone always ends up doing symlinks
2. By default, non-administrative users in Windows don’t have permissions to symlink directories

So, you’ll need to run as admin when installing with npm 3 on Windows. 

[...]

Issue #429, opened 2015-06-26
npm install error


marklague:
I tried installing intern both locally and globally and I've been getting this error 

>C:\Users\mlague\Documents\NetBeansProjects>npm install intern
>npm WARN excluding symbolic link query -> vendors/jquery/src/
>npm WARN excluding symbolic link query -> vendors/jquery/src/
>
>intern@2.2.2 install C:\Users\mlague\Documents\NetBeansProjects\node_modules\intern
>node support/fixdeps.js
>
>fs.js:717
>  return binding.symlink(preprocessSymlinkDestination(destination, type),
>                 ^
>Error: EPERM, operation not permitted 'c:\Users\mlague\Documents\NetBeansProjects\node_modules\intern\node_modules\dojo'
>    at Object.fs.symlinkSync (fs.js:717:18)
>    at C:\Users\mlague\Documents\NetBeansProjects\node_modules\intern\support\fixdeps.js:33:6
>    at Array.forEach (native)
>    at Object.<anonymous> >(C:\Users\mlague\Documents\NetBeansProjects\node_modules\intern\support\fixdeps.js:14:28)
>    at Module._compile (module.js:456:26)
>    at Object.Module._extensions..js (module.js:474:10)
>   at Module.load (module.js:356:32)
>    at Function.Module._load (module.js:312:12)
>    at Function.Module.runMain (module.js:497:10)
>    at startup (node.js:119:16)
>npm ERR! Windows_NT 6.1.7601
>npm ERR! argv "C:\\webApps\\nodeJS\\\\node.exe" "C:\\webApps\\nodeJS\\node_modules\\npm\\bin\\npm-cli.js" "install" "intern"
>npm ERR! node v0.11.0
>npm ERR! npm  v2.11.1
>npm ERR! code ELIFECYCLE
>
>npm ERR! intern@2.2.2 install: `node support/fixdeps.js`
>npm ERR! Exit status 8
>npm ERR!
>npm ERR! Failed at the intern@2.2.2 install script 'node support/fixdeps.js'.
>npm ERR! This is most likely a problem with the intern package,
>npm ERR! not with npm itself.
>npm ERR! Tell the author that this fails on your system:
>npm ERR!     node support/fixdeps.js
>npm ERR! You can get their info via:
>npm ERR!     npm owner ls intern
>npm ERR! There is likely additional logging output above.
>
>npm ERR! Please include the following file with any support request:
[...]

PR #501, opened 2015-09-22
Proposed fix for #500 and #429

gwicksted:
Resolves relative path `'..'` and possible mixed casing of `C:\` vs `c:\` on Windows due to `__dirname`.

### DIFF
--- a/support/fixdeps.js
+++ b/support/fixdeps.js
@@ -1,7 +1,7 @@
 /*jshint node:true */
 var fs = require('fs');
 var path = require('path');
-var expected = path.join(__dirname, '..', 'node_modules');
+var expected = path.normalize(path.join(__dirname, '..', 'node_modules'));
 
 // Create the node_modules directory if it doesn't yet exist, such as when all of
 // Intern's dependencies were already installed by a parent package.
@@ -27,7 +27,7 @@ if (!fs.existsSync(expected)) {
 		}
 	}
 
-	var actualPath = path.dirname(require.resolve(path.join(dependency, 'package.json')));
+	var actualPath = path.normalize(path.dirname(require.resolve(path.join(dependency, 'package.json'))));
 
 	if (actualPath.indexOf(expectedPath) !== 0) {
 		fs.symlinkSync(actualPath, expectedPath, 'dir');


gwicksted:
Travis CI build failed due to IE timeout (unrelated to this fix)

gwicksted:
Closing PR in hopes of triggering a rebuild with Travis CI.

gwicksted:
Reopen pull request to trigger Travis CI.

gwicksted:
Sigh... another unrelated Travis CI error.  Please accept the PR!

csnover:
Thanks for the patch, do you have an upstream bug report about this problem? It seems wrong that `require.resolve` and `__dirname` would provide paths with different cases on the drive letter.

gwicksted:
I wish I did ... but I think they've rejected all previous attempts to fix related issues.  Instead it remains not well documented but a "generally accepted practice" when performing string comparison on two paths especially if they might exist on a case-insensitive file system (ie. NTFS/FAT on Windows).

It has some pretty scary side effects when loading modules using `"./"` vs `__dirname` and the module cache.

One of the better tickets:

[require() loads same module multiple times if case is different](https://github.com/nodejs/node-v0.x-archive/issues/6829)

Ticket that specifically outlines this issue (about half way down) when loading modules:

[https://github.com/nodejs/node-v0.x-archive/issues/7031](https://github.com/nodejs/node-v0.x-archive/issues/7031)

csnover:
Thanks for the pointers. I see the commit that was added to Node.js that broke this between 0.10 and 0.12. (Why would you ever normalise drive letters to lowercase? When are drive letters ever represented lowercase anywhere ever? Uppercase drive letters date back to at least CP/M in the 1970s. How cou… I jus… whatever. \</rant>)

I’m surprised this isn’t more of an issue more frequently.

gwicksted:
That about sums it up!  I suppose nodejs is less frequently run on Windows and those of us who have been burned by this bug just instinctively normalize ... at least it hasn't reached the point of the [double claw hammer](http://eev.ee/blog/2012/04/09/php-a-fractal-of-bad-design/).  ;)

csnover:
So sorry, can you send a test that causes this to actually happen reliably outside of the context of npm 3? I am trying to get `__dirname` to not be what I expect it to be on Windows in Node 4.1.1 and I don’t seem to be able to do so. I also don’t see the bad code from 0.11 any more in lib/path.js and can’t seem to find any equivalent code in current Node.js master or 4.x branches that would break the drive letter.

I see an installation failure when using npm 3, but only npm 3. npm 3 failure is probably due to the fact that npm 3 does everything in some `.staging` directory and then moves everything once the installs were “successful”, not thinking that, you know, someone might need the dependencies to actually exist in their final locations before running postinstall scripts.

csnover:
However… npm 3 actually seems to work OK on Linux and the deps seem to be copied over *before* the install scripts run in that case so actually I am not sure what is the deal. I would really appreciate stronger verification on what is going on. :)

csnover:
OK, so the issue here is:

1. npm 3 doesn’t install dependencies into the `intern/node_modules` directory so Windows users always have to do the symlinking
2. By default, non-administrative users in Windows don’t have permissions to symlink.

So, you need to run as admin when installing with npm 3 on Windows. The fixdeps command can also optionally be updated to perform copies when a symlink fails with an EPERM error, if anyone cares to make that modification.

csnover:
Or actually junction links seem to work fine without any permissions changes, so maybe there is a way to have fixdeps do that instead.

gwicksted:
Very good!  Upon further testing you're absolutely right -- it seems they did fix that bug!

PR #519, opened 2015-10-08
separate intern's Proxy and dig dug tunnel proxy settings

### LINKED ISSUE #509: Separate proxy settings for dig dug tunnel and intern Proxy server
Currently intern Proxy server seems to be using the same proxy settings as tunnel (defined in tunnelOptions). This is not always wanted behavior - i.e. I would like to specify different proxy settings for my tunnel and different for intern Proxy server (I'd like it to not use proxy at all).

bartoszkaczorek:
Fixes #509

### DIFF
--- a/lib/executors/Runner.js
+++ b/lib/executors/Runner.js
@@ -162,6 +162,14 @@ define([
 				.then(startTunnel);
 		},
 
+		_buildServerOptions : function(config) {
+			var result = {};
+			if (config.internServerOptions !== undefined && config.internServerOptions.proxy !== undefined) {
+				result.proxy = config.proxy;
+			}
+			return result;
+		},
+
 		/**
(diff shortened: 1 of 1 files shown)


dylans:
@jason0x43 we should probably make a decision on this one, given that @bartoszkaczorek has been waiting a while on this PR.

bartoszkaczorek:
Any update on that one?

dylans:
@bartoszkaczorek thanks for making the standalone module... we'll try to land all of this for Intern 4, though we're unlikely to look at it until our TS refactor is finished.

jason0x43:
This functionality was added in a similar way in https://github.com/theintern/intern/commit/71c22829b1b51de0d2b6ba38ee7eda4ec9ec9cbc.

PR #520, opened 2015-10-18
Use relative symlinks on POSIX, for reliability.

### LINKED ISSUE #510: Re-naming user's project breaks client.html (symlinks)
**Environment:**
Intern: `3.`
Node: `4.1.1`
npm: `3.3.5`

**Steps:**
1. Setup a typical project that depends on Intern, as recommended in the [user guide](https://theintern.github.io/intern/#directory-structure "Recommended directory structure for projects that use Intern.").
2. Verify that the [browser client](https://theintern.github.io/intern/#browser-client) works.
3. Re-name the project directory for the app that depends upon Intern.
4. Try to use the browser client again (with the appropriately updated URL).

**Current behavior:**
The browser client fails to load, due to out-of-date symlinks. Assuming you renamed the project directory to `my-fav-project`, you receive a 403 error for this:
```
http://localhost/my-fav-project/node_modules/intern/node_modules/dojo/loader.js 
```
That happens because the symlink at ...
```
my-fav-project/node_modules/intern/node_modules/dojo
```
... points to the no-longer-existing ...
```
my-project/node_modules/dojo
```

This issue crops up if any path component leading up to the project directory is renamed, it's not just the directory name itself, but the full path, because the symlink is absolute (even on POSIX systems).

**Expected behavior:**
Intern users are able to be blissfully unaware of any symlink issues and can change their project's path, as happens from time to time, without breaking Intern's features such as the browser client. 

**Suggested fix:**
Change [this line](https://github.com/theintern/intern/blob/master/client.html#L18) at the top of client.html (and elsewhere such changes are needed, such as below) from:
```html
<script src="node_modules/dojo/loader.js"></script>
```
... to ...
```html
<script src="../dojo/loader.js"></script>
```

and its [map config](https://github.com/theintern/intern/blob/master/client.html#L48) to:
```
map: {
    intern: {
        dojo: 'intern/../dojo',
        chai: 'intern/../chai/chai',
        diff: 'intern/../diff/diff'
    },
    '*': {
        'intern/dojo': 'intern/../dojo'
    }
}
```

Now, since we are no longer relying upon symlinks, renaming is fine. Although I'm not familiar enough with the internals to know for sure whether this will cause issues somewhere. For installs with npm@2, symlinks will have to be used, which seems reasonable as a gracefully degrading UX for older environments.

sholladay:
This changes the symlinks for Intern's AMD dependencies from absolute to relative, so that changes in the dependent project's path (such as re-naming) will not break them. Fixes #510.

### DIFF
--- a/support/fixdeps.js
+++ b/support/fixdeps.js
@@ -30,6 +30,6 @@ if (!fs.existsSync(expected)) {
 	var actualPath = path.dirname(require.resolve(path.join(dependency, 'package.json')));
 
 	if (actualPath.indexOf(expectedPath) !== 0) {
-		fs.symlinkSync(actualPath, expectedPath, 'junction');
+		fs.symlinkSync(path.relative(path.dirname(expectedPath), actualPath), expectedPath, 'junction');
 	}
 });


csnover:
Thank you, all patches are greatly appreciated. Sorry for the delay landing.

PR #523, opened 2015-10-30
waitForRunner updates

jason0x43:
- `lib/Proxy.js` now looks for waitForRunner in the correct location
  (self.config rather than self).
- `lib/Proxy.js` now passes the message payload (which is what
  contains the event name) to getShouldWait
- Allow waitForRunner to be an array specifying which events to wait
  for. The existing behavior for values of `true` and 'fail' is
  unchanged.


csnover:
Thanks!

PR #524, opened 2015-11-04
Fall back to copying things to node_modules/ if symlinking fails.

Arnavion:
Symlinking fails on a Windows machine when installing with npm 3.x where the intern install directory is on a Samba share (EUNKNOWN, the underlying win32 error is "NOT REPARSE POINT" from FileSystemControl). @csnover mentions in https://github.com/theintern/intern/issues/500#issuecomment-142804696 that falling back to copying is an acceptable option if symlinking fails.

Uninstalling with `npm uninstall` appears to delete the intern/node_modules directory and the intern/ directory, so there's no need for a separate uninstall script to delete it manually.

---

With this change:

- Installing on Linux with target directory on Linux (no change): symlink `/full/path/node_modules/intern/node_modules/chai` pointing to `../../chai`

- Installing on Windows with target directory on Windows (no change): ~~symlink~~ junction point `C:\full\path\node_modules\intern\node_modules\chai` pointing to `C:\full\path\node_modules\chai`

- Installing on Windows with target directory on Samba share: Directory `C:\full\path\node_modules\intern\node_modules\chai` with contents of `C:\full\path\node_modules\chai`

### DIFF
--- a/support/fixdeps.js
+++ b/support/fixdeps.js
@@ -30,6 +30,43 @@ if (!fs.existsSync(expected)) {
 	var actualPath = path.dirname(require.resolve(path.join(dependency, 'package.json')));
 
 	if (actualPath.indexOf(expectedPath) !== 0) {
-		fs.symlinkSync(path.relative(path.dirname(expectedPath), actualPath), expectedPath, 'junction');
+		try {
+			fs.symlinkSync(path.relative(path.dirname(expectedPath), actualPath), expectedPath, 'junction');
+		}
+		catch (error) {
+			console.warn('Symlinking %s to %s failed with %s. Copying instead...', actualPath, expectedPath, error.code);
+			copy(actualPath, expectedPath);
+		}
 	}
 });
+
+function copy(source, target) {
+	try {
+		var stats = fs.statSync(source);
+	}
+	catch (error) {
+		if (error.code !== 'ENOENT') {
+			throw error;
+		}
+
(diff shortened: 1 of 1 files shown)


support/fixdeps.js
```diff
@@ -30,6 +30,43 @@ if (!fs.existsSync(expected)) {
...
+
+		return;
+	}
+
+	if (stats.isDirectory()) {
+		try {
+			fs.mkdirSync(target);
+		}
+		catch (error) {
+			if (error.code !== 'EEXIST') {
```
Arnavion: This isn't strictly necessary, since even reinstalling will uninstall first (and thus delete the node_modules directory), so this directory cannot exist.

support/fixdeps.js
```diff
@@ -30,6 +30,43 @@ if (!fs.existsSync(expected)) {
...
+			fs.symlinkSync(path.relative(path.dirname(expectedPath), actualPath), expectedPath, 'junction');
+		}
+		catch (error) {
+			console.warn('Symlinking %s to %s failed with %s. Copying instead...', actualPath, expectedPath, error.code);
+			copy(actualPath, expectedPath);
+		}
 	}
 });
+
+function copy(source, target) {
```
Arnavion: Change this to use ncp or other such dependency instead of hand-rolling?

csnover:
Thanks!

PR #530, opened 2015-11-12
Fix the missing suiteError in Pretty report mode

LingSpb:
Mark suite as FAIL when we got suiteError + fix the error message to include the suite id


csnover:
Thanks for the patch!

PR #534, opened 2015-11-18
Inline source map

kitsonk:
This adds the feature that detects if a sourceMap is included inline in the file and properly decodes them to remap back.

Also, I believe I addressed a logic error in the code.  On [lib/util.js:298](https://github.com/theintern/intern/blob/aff8eb9b01bf35335a2c2a6c4f1df7b6af527375/lib/util.js#L298) the code tried to detect the existence of the map in the hash based on the `filepath` but whenever the map was [set in the hash](https://github.com/theintern/intern/blob/aff8eb9b01bf35335a2c2a6c4f1df7b6af527375/lib/util.js#L319) it used instead the filename of the map file.  This resulted in the first block of code never being executed (and a fairly inefficient situation where you would always parse the name of the mapfile.

Since, obviously, there is no filename for inline sources, I have changed the `fileSourcesMaps` to be indexed on `filepath` this also then eliminated another if statement where the hash was being checked again if the file was in there.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -1,3 +1,4 @@
+/* jshint node: true */
 define([
 	'dojo/has',
 	'dojo/lang',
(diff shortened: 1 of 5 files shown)


csnover:
Thanks!

PR #426, opened 2015-06-24
Add support for Istanbul Json Coverage

kitsonk:
This add the support for allowing the generation of the Istanbul Json Coverage mentioned in #425.

PR #545, opened 2015-11-28
Add result parser to grunt task

JSMike:
Adds 'resultParser' option for the intern grunt task.
resultParser is a function that takes in the result from spawning istanbul. The function is called after the istanbul task has completed without errors. The function should return null if there are no issues with the results, or return an Error if the results aren't as expected.

Here's an example use case to enforce an overall coverage percentage:
```javascript
var testThreshold = 70;
/**
 * @name resultParser
 * @description checks results from istanbul to confirm expected results
 * @param { object } result, result from gulp.util.spawn callback
 * @returns { error } if unexpected result
 * @returns { null } if result is expected
 **/
var resultParser = function (result) {
    // ansi colors regexp from http://stackoverflow.com/a/29497680
    var colors = /[\u001b\u009b][\[()#;?]*(?:[0-9]{1,4}(?:;[0-9]{0,4})*)?[0-9A-ORZcf-nqry=><]/g,
        // Get total coverage line from istanbul tables
        coverage = result.stdout
            .replace(colors, '')
            .split('\n')
            .filter(function (el) {
                return el.indexOf('All files') > -1;
            });
    // Get each coverage type's column value from the last match
    coverage = coverage[coverage.length - 1]
        .split('|')
        .map(function (el) {
            return !isNaN(parseFloat(el.trim())) ? parseFloat(el.trim()) : null;
        })
        .filter(function (el) {
            return el !== null;
        });
    // Get the average coverage
    coverage = coverage.reduce(function (total, next) {
        return total + next;
    }) / coverage.length;
    // Compare coverage to threshold
    if (coverage < testThreshold) {
        grunt.log.error('Coverage: ' + coverage + '%');
        return new Error('Coverage below ' + testThreshold + '% threshold!');
    } else  {
        grunt.log.ok('Coverage: ' + coverage + '%');
    }
    return null;
};

grunt.initConfig({
    intern: {
        all: {
            options: {
                runType: 'runner', // defaults to 'client'a
                config: 'tests/intern',
                resultParser: resultParser
            }
        }
    }
});
```

### DIFF
--- a/tasks/intern.js
+++ b/tasks/intern.js
@@ -64,7 +64,8 @@ module.exports = function (grunt) {
 				sauceUsername: true,
 				testingbotKey: true,
 				testingbotSecret: true,
-				nodeEnv: true
+				nodeEnv: true,
+				resultParser: true
 			};
 
 		Object.keys(opts).forEach(function (option) {
@@ -96,7 +97,8 @@ module.exports = function (grunt) {
 			'sauceUsername',
 			'testingbotKey',
 			'testingbotSecret',
-			'nodeEnv'
+			'nodeEnv',
+			'resultParser'
 		].forEach(function (option) {
 			var value = opts[option];
 			if (value) {
@@ -114,10 +116,18 @@ module.exports = function (grunt) {
 				cwd: process.cwd(),
 				env: env
 			}
-		}, function (error) {
-			// The error object from grunt.util.spawn contains information
-			// that we already logged, so hide it from the user
-			done(error ? new Error('Test failure; check output above for details.') : null);
+		}, function (error, result) {
+			// The stdout and stderr of grunt.util.spawn are already being logged
+			// If error then return an error object to grunt
+			if (error) {
+				return done(new Error('Test failure; check output above for details.'));
+			}
+			// If a result parser function was passed in return result of function call
+			if (typeof env.RESULT_PARSER === 'function') {
+				return done(env.RESULT_PARSER(result));
+			}
+			// Return grunt callback with no error
+			return done(null);
 		});
 
 		child.stdout.on('data', readOutput);


JSMike:
Note: There are currently 0 tests for the grunt task.

csnover:
I feel weird about the level of output parsing. #199 was created partly because it provides an easy-to-use machine-readable format that’s guaranteed not to change that can be used to exchange messages from Intern to Grunt. There’s no guarantee with the text output of Istanbul that it won’t change to some other human-readable representation in the future.

I’d rather add a new configuration to Intern itself that will make it fail the test run if code coverage is below a certain, configurable amount. Would you be willing/able to do that instead? I thought there was an open ticket for such an enhancement but I can’t seem to find one.

kitsonk:
Just thinking out loud... would a better pattern, instead of convoluting the Intern config, is create a test interface that exposes some of these internals that an end user could then choose to pass or throw an error?  Then it would be up to the end user to have this final test case that throws under whatever arbitrary conditions it has or hasn't met?  For example something like:

```javascript
declare([ 'intern!object' ], function (registerSuite) {

  registerSuite({
    name: 'the final stand',
    'coverage threshold': function () {
      if (this.coverage.branch < 0.90) {
        throw new Error('Not enough coverage, try again!');
      }
    }
  });

});
```

JSMike:
I agree it would be better to just have a configurable passing threshold.
There's a PR for adding a coverage.json reporter https://github.com/theintern/intern/pull/426, this should be compatible with https://www.npmjs.com/package/grunt-istanbul-coverage.
It would be nice if there was just configuration for a passing threshold in intern, whether it be a for each aspect like the grunt task I linked or an average of all.

JSMike:
@csnover I'll take a look into a better solution for this. My initial thought was to give developers access to the spawn result so they can manipulate it as necessary, this was just one application of how it could be used. But I agree it's not the best way to get coverage.

Note: I'm very busy through the end of the week, but next week I'll be able to look into this some more.

JSMike:
Closing this as I've come up with a different solution which I'll be posting separately.

Issue #551, opened 2015-12-04
Addressing a source of false positive tests


wonnage:
I like being able to return a promise instead of using `this.async`, but it's easy to forget that return statement, and now your test is immediately marked as succeeded (while the actions continue on in the background, as the following tests run in the same session...)

Would be great if test success was made more explicit - maybe usage of `this.remote` ought to mark the test as async as well? I've seen other test frameworks pass a done callback into the test functions.

csnover:
It would be possible to mark a test as async if someone tries to access `remote`, but no way to correctly handle the actual usage of the remote object, since it generates a Promise object chain rather than being a single object with an internally held stack or queue.

Maybe it would be better to flag any access of `remote`, check if the test is async after the test function runs, and if remote was access and there’s no async handling, emit a warning? That seems like it strikes the right balance between not insane to program and helpful to end-users. Let me know what you think.

With regards to the done-callback-signature, I don’t feel that is the right approach generally, but of course if you really think this would help you/your team you can write a custom test interface that has something like that too, and if that is super popular for some reason then it could be included I guess.

wonnage:
I like the warning idea, does the approach in the patch look reasonable?

csnover:
Yep, it’s on the right track! I sent some line comments.

sholladay:
My second favorite test framework, [ava](https://github.com/sindresorhus/ava), has the concept of an [assertion plan](https://github.com/sindresorhus/ava#assertion-plan) that helps with this sort of thing a ton. In fact, it allows a test to be considered failed until proven otherwise.

csnover:
Assertion count check is something that already exists in the QUnit test interface today. However I would think if someone forgets to return a promise they will just as easily forget to call a method for counting assertions too?

As an aside, the reason why there is no fully integrated assertion count now (i.e. why it is only in the QUnit interface) is because it’s not possible to count number of assertion checks if someone is using their own assertion library, so it would either be some extra separate check-pass API, or we need to expose some thing to let people “bless” their preferred assertion library, or we would need to restrict what assertion library people use when they also want to use assertion count check.

sholladay:
Make it so a test must explicitly opt-in or opt-out of having an assertion plan (or less drastically, just a config flag to opt-in globally). Since a test must end having made exactly the number of assertions specified in the plan count, if you forget anything, the test fails with a clear message (too many or too few assertions or no plan).

[...]

PR #552, opened 2015-12-04
warn if remote is accessed without isAsync

wonnage:
re: #551

### DIFF
--- a/lib/Test.js
+++ b/lib/Test.js
@@ -41,6 +41,7 @@ define([
 		 * @see Suite#remote
 		 */
 		get remote() {
+			this._usesRemote = true;
 			return this.parent.remote;
 		},
 
@@ -256,6 +257,10 @@ define([
 				})
 				.then(function () {
 					self.hasPassed = true;
+					if (self._usesRemote && !self.isAsync) {
+						console.warn("This test uses the remote but isn't marked as async! Intern won't wait for your test actions to complete.");
+						console.warn("You probably want to call this.async() or return a promise.");
+					}
 					return report('testPass');
 				}, function (error) {
 					if (error === SKIP) {
--- a/lib/Test.js
+++ b/lib/Test.js
@@ -18,6 +18,7 @@ define([
 		parent: null,
 		isAsync: false,
 		timeElapsed: null,
+		_usesRemote: false,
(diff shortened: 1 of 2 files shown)


lib/Test.js
```diff
@@ -41,6 +41,7 @@ define([
 		 * @see Suite#remote
 		 */
 		get remote() {
+			this._usesRemote = true;
```
csnover: The flag needs to be reset when the test is reset. It should also be listed explicitly in the prototype with its default value.

lib/Test.js
```diff
@@ -256,6 +257,10 @@ define([
 				})
 				.then(function () {
 					self.hasPassed = true;
+					if (self._usesRemote && !self.isAsync) {
+						console.warn("This test uses the remote but isn't marked as async! Intern won't wait for your test actions to complete.");
```
csnover: Messages should go through `reporterManager` so the user can handle and filter them, rather than sending messages directly to console.

wonnage: So should this just be a this.report('warnAsync') or something? I'm not sure how that part of the code works.

wonnage:
Updated with @csnover feedback; this now marks a test as skipped if you use this.remote without marking the test as async, with the warning as the skip reason. 

I was struggling to think of a case where you'd want a synchronous test while using a remote web driver and couldn't; hopefully we can agree that this is never useful.

jason0x43:
Rather than emitting a warning when remote is used without setting async, it might make more sense to throw an exception and fail the test rather than skipping it. I mean, if you're doing something nonsensical, the testing system might as well make it super obvious.

dylans:
Thanks for the PR @wonnage ... did you get a chance to address @jason0x43's feedback?

wonnage:
Hey sorry, I forgot I had this PR out there. When I submitted it, the company I was working at had a bunch of tests "passing" due to this. The upgrade process would be easier/allow incremental fixing if intern gave a warning rather than just failing. 

I no longer work for them, so I'm open to making it throw if that gets this merged, though!

jason0x43:
I went ahead and changed the skip to a failure and updated the unit tests. This has been merged into master. Thanks!

PR #553, opened 2015-12-11
Add GruntCoverage reporter

JSMike:
Hello,

I had previously raised a pull request to allow a function to parse stdout https://github.com/theintern/intern/pull/545. @csnover mentioned "I’d rather add a new configuration to Intern itself that will make it fail the test run if code coverage is below a certain, configurable amount."

In response I've created a reporter that takes in the coverage results and compares them against threshold config that's passed in.

This reporter requires a threshold config object containing the keys of the coverage types and threshold value to compare against. In the example below shows all possible threshold types, including an 'overall' option that will compare the average of the four primary coverage types against the passed value. None of the threshold options are mandatory, the end user can include any combination of coverage options. If no threshold config is passed in it defaults to `{ overall: 75 }`. Only the coverage options included in the config will be enforced and reported on.

Example reporters config:
```
    reporters: [
        "Runner",
        {
            id: "GruntCoverage",
            threshold: {
                statements: 75,
                branches: 65,
                functions: 75,
                lines: 75,
                overall: 75
            }
        }
    ]
```

Example output:
```
-----------------------------------------------------------|----------|----------|----------|----------|----------------|
All files                                                  |    75.89 |    68.73 |    76.75 |    76.07 |                |
-----------------------------------------------------------|----------|----------|----------|----------|----------------|

>> TOTAL: tested 1 platforms, 0/589 tests failed
>> PASSED! Lines: 76.07%, above threshold of 75%
>> PASSED! Statements: 75.89%, above threshold of 75%
>> PASSED! Functions: 76.75%, above threshold of 75%
>> PASSED! Branches: 68.73%, above threshold of 65%
>> FAILED! Overall: 74.36%, below threshold of 75%
Warning: Coverage threshold not met! Use --force to continue.
```
**Note:** This reporter requires grunt, which isn't currently a dependency for Intern, even though Intern includes a grunt task. I'm unable to add tests for this reporters because grunt won't be available, unless it's desired to be added. Since I was able to achieve this functionality by only using a reporter I'm not sure if it's desired for this reporter to be included with intern, because of the grunt dependency, or if it should be a separate external/custom reporter that can be added in optionally.

@csnover let me know if this is what you had in mind, or if this should go in another direction that doesn't require grunt. I made this require grunt because I was intending to use it in conjunction with the intern grunt task.

I went in this direction, using a reporter to enforce coverage, because the reporters always receive the coverage and runEnd events, but I don't see a clear way to get reporters to cause the spawned intern task to report a failure without causing a stack trace, thats why I incorporated grunt to have that control.

### DIFF
--- a/lib/ProxiedSession.js
+++ b/lib/ProxiedSession.js
@@ -78,21 +78,29 @@ define([
 			}
 
 			if (this.coverageEnabled) {
+				var promise;
+
 				// At least Safari will not inject user scripts for non http/https URLs, so we can't get coverage data.
 				if (this.capabilities.brokenExecuteForNonHttpUrl) {
-					return this.getCurrentUrl().then(function (url) {
-						if ((/^https?:/i).test(url)) {
-							return self.execute(getCoverageData, [ self.coverageVariable ])
-								.then(function (coverageData) {
-									return coverageData &&
-									self.reporterManager.emit('coverage', self.sessionId, JSON.parse(coverageData));
-								}).finally(function () {
-									return _super.get.apply(self, args);
-								});
-						}
-						return _super.get.apply(self, args);
+					promise = this.getCurrentUrl().then(function (url) {
+						return (/^https?:/i).test(url);
 					});
 				}
+				else {
+					promise = Promise.resolve(true);
+				}
+
+				return promise.then(function (shouldGetCoverage) {
+					if (shouldGetCoverage) {
+						return self.execute(getCoverageData, [ self.coverageVariable ])
+							.then(function (coverageData) {
+								return coverageData &&
+								self.reporterManager.emit('coverage', self.sessionId, JSON.parse(coverageData));
+							});
+					}
+				}).finally(function () {
+					return _super.get.apply(self, args);
+				});
 			}
 
 			return _super.get.apply(self, args);
--- a/lib/ProxiedSession.js
+++ b/lib/ProxiedSession.js
@@ -78,17 +78,24 @@ define([
 			}
 
 			if (this.coverageEnabled) {
-				return this.execute(getCoverageData, [ this.coverageVariable ])
-					.then(function (coverageData) {
-						return coverageData &&
-							self.reporterManager.emit('coverage', self.sessionId, JSON.parse(coverageData));
-					}).finally(function () {
+				// At least Safari will not inject user scripts for non http/https URLs, so we can't get coverage data.
+				if (this.capabilities.brokenExecuteForNonHttpUrl) {
+					return this.getCurrentUrl().then(function (url) {
+						if ((/^https?:/i).test(url)) {
+							return self.execute(getCoverageData, [ self.coverageVariable ])
+								.then(function (coverageData) {
+									return coverageData &&
+									self.reporterManager.emit('coverage', self.sessionId, JSON.parse(coverageData));
+								}).finally(function () {
+									return _super.get.apply(self, args);
+								});
+						}
 						return _super.get.apply(self, args);
 					});
+				}
 			}
-			else {
-				return _super.get.apply(self, args);
-			}
+
+			return _super.get.apply(self, args);
 		},
 
 		/**
--- a/lib/reporters/CoverageThreshold.js
+++ b/lib/reporters/CoverageThreshold.js
@@ -1,0 +1,73 @@
+define([
+    'dojo/node!istanbul/lib/collector',
+    'dojo/node!istanbul/lib/object-utils',
+    'dojo/node!istanbul/lib/report/common/defaults'
+], function (Collector, utils, defaults) {
+    var logger;
+    function CoverageThreshold(config) {
+        var defaultThresholds = defaults.watermarks();
+
+        for (var key in defaultThresholds) {
+            defaultThresholds[key] = defaultThresholds[key][0];
+        }
+
+        this.config = config || {};
+        logger = this.config.logger ? this.config.logger : console;
+
+        config.threshold = config.threshold || defaultThresholds;
+        this._collector = new Collector();
+    }
+
+    function enforceThreshold(threshold, actual, expected) {
+        var name = threshold[0].toUpperCase() + threshold.slice(1);
+        var success = actual >= expected;
+        var placing = success ? 'above' : 'below';
+        var passed = success ? 'PASSED! ' : 'FAILED! ';
+        var status = success ? 'log' : 'error';
+        var msg = passed + name + ': ' + actual + '%, ' + placing + ' threshold of ' + expected + '%';
+
+        logger[status](msg);
+        return success;
+    }
+
+    CoverageThreshold.prototype.coverage = function (sessionId, coverage) {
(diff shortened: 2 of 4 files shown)


lib/reporters/GruntCoverage.js
```diff
@@ -0,0 +1,64 @@
...
+        });
+
+        finalSummary = utils.mergeSummaryObjects.apply(null, summaries);
+
+        for (var key in finalSummary) {
+            if (finalSummary[key].pct) {
+                if (this.config.threshold.hasOwnProperty(key)) {
+                    success = success && enforceThreshold(key, finalSummary[key].pct, this.config.threshold[key]);
+                }
+                overall += finalSummary[key].pct;
```
csnover: I don’t think you can just combine function, statement, line, and branch coverage and then divide to get an average of like this since they are measures of different incompatible units.

csnover:
> I don't see a clear way to get reporters to cause the spawned intern task to report a failure without causing a stack trace, thats why I incorporated grunt to have that control.

The trace on simple test run failure is a problem that can be addressed separately (the easiest way would probably be to set a flag on the error object that causes it to not be emitted through `emitFatalError`); don’t let it stop you from implementing *without* Grunt by making these changes:

1. Send messages to the user using the `console` object passed to the reporter constructor instead of `grunt.log`.
2. To cause the executor to exit with a non-zero status: for now, I would recommend just exposing an extra property on `lib/Executor` and setting it through `runEnd` of this reporter (this event receives the executor as its first argument). Then if this property is true, fail the test run, [just like this code does for suite errors](https://github.com/theintern/intern/blob/master/lib/executors/Executor.js#L300-L302).

Eventually, it makes sense for the reporter system to be expanded to be a general mechanism for hooks into the test system, so I’m OK with having this functionality in a reporter, even though it seems a little weird to influence the result of the test system through a ‘reporter’. :)

Let me know if this makes sense or you can think of some reason why this is a bad approach.

lib/reporters/GruntCoverage.js
```diff
@@ -0,0 +1,64 @@
...
+    'dojo/node!grunt'
+], function (Collector, utils, grunt) {
+    function GruntCoverage(config) {
+        this.config = config || {};
+        config.threshold = config.threshold || { overall: 75 };
+        this._collector = new Collector();
+    }
+
+    function enforceThreshold(threshold, actual, expected) {
+        var name = threshold[0].toUpperCase() + threshold.slice(1),
```
csnover: One var per var per code convention, thanks!

lib/reporters/GruntCoverage.js
```diff
@@ -0,0 +1,64 @@
...
+        grunt.log[status](msg);
+        return success;
+    }
+
+    GruntCoverage.prototype.coverage = function (sessionId, coverage) {
+        this._collector.add(coverage);
+    };
+
+    GruntCoverage.prototype.runEnd = function () {
+        var collector = this._collector,
```
csnover: One var per var per code convention, thanks!

lib/reporters/GruntCoverage.js
```diff
@@ -0,0 +1,64 @@
+define([
+    'dojo/node!istanbul/lib/collector',
+    'dojo/node!istanbul/lib/object-utils',
+    'dojo/node!grunt'
+], function (Collector, utils, grunt) {
+    function GruntCoverage(config) {
```
csnover: Probably want to rename this to something like “CoverageEnforcementReporter” or something similar but better than what I can come up with :).

lib/reporters/GruntCoverage.js
```diff
@@ -0,0 +1,64 @@
...
+        this.config = config || {};
+        config.threshold = config.threshold || { overall: 75 };
```
csnover: I’d probably recommend getting defaults from `istanbul/lib/report/common/defaults.watermarks()` and use the low watermark as the default for failure, for consistency.

lib/reporters/GruntCoverage.js
```diff
@@ -0,0 +1,64 @@
...
+
+        collector.files().forEach(function (file) {
+            summaries.push(utils.summarizeFileCoverage(collector.fileCoverageFor(file)));
+        });
+
+        finalSummary = utils.mergeSummaryObjects.apply(null, summaries);
+
+        for (var key in finalSummary) {
+            if (finalSummary[key].pct) {
+                if (this.config.threshold.hasOwnProperty(key)) {
```
csnover: Use `key in this.config.threshold` instead of `hasOwnProperty` or else this breaks anyone that uses `Object.create` or a constructor with prototype.

lib/reporters/GruntCoverage.js
```diff
@@ -0,0 +1,64 @@
...
+                    success = success && enforceThreshold(key, finalSummary[key].pct, this.config.threshold[key]);
```
csnover: If an earlier threshold fails this will short circuit and not run `enforceThreshold` for this `key`; is that really what is intended?

csnover:
That’s all I’ve got. Thanks for sending this! Let me know if you have any questions on the feedback.

JSMike:
@csnover thanks for the feedback. I'll work on implementing the suggested changes. I'll hop on IRC or msg here if I have questions.

JSMike:
Hi All, I just wanted to note that I'm still looking into this, I just have a bunch of other things I'm working on that are higher priority for me. I'll try to get this prioritized sooner than later.

dylans:
@JSMike just checking in to see if you expect to have time for this in the near future. No rush, reaching out.

JSMike:
Hi @dylans 

Yes, I'm still here. I've been busy with full time job and half-time school. I'll see if I can get some time to work on this soon.

JSMike:
FYI, this is covered under a corporate CLA.

lib/reporters/GruntCoverage.js
JSMike: changed to "CoverageThreshold"

lib/reporters/GruntCoverage.js
JSMike: Fixed.

lib/reporters/GruntCoverage.js
JSMike: Fixed.

lib/reporters/GruntCoverage.js
JSMike: done.

lib/reporters/GruntCoverage.js
JSMike: done

lib/reporters/GruntCoverage.js
JSMike: If any threshold fails we'd want to halt for CI.

JSMike:
Only thing to mention is that i left in the 'overall' option, which averages the 4 columns. I know that @csnover mentioned that these columns measure different things, but for our project's CI we just wanted to enforce the average of the numbers instead of setting a specific threshold for each type of coverage.

PR #557, opened 2015-12-18
Update Chai to v3.4.1.

mwistrand:
Updates Intern to use the latest version of Chai available on npm (v3.4.1), and includes updates to the corresponding `chai.d.ts`.


jason0x43:
chai was updated to 3.5.0 in fb0791f, and the typings were updated to the latest (for 3.4.0) in 6b21d22.

PR #559, opened 2015-12-22
Handle globs in suite lists

### LINKED ISSUE #359: Auto-glob in Node.js

jason0x43:
Glob entries of the form 'package/dir/**/*' are now resolved to lists of module IDs. In a Node.js context, Intern will resolve the package name to its filesystem path and use the `glob` package to find all matching modules. These modules will be converted back to module IDs and added to the Intern config. Both `suites` and `functionalSuites` are handled this way.

In the browser context, the `suites` list is now always sent back to a special `__resolveSuites__` endpoint on Intern's proxy for resolution. The list is resolved in the same manner as in the Node client, and the resolved suites list is returned to the browser and added to Intern's config.

Resolution happens in `PreExecutor` just after loader swapping, so loader maps are in effect

Fixes #359

### DIFF
--- a/lib/Proxy.js
+++ b/lib/Proxy.js
@@ -6,9 +6,10 @@ define([
 	'dojo/node!path',
 	'dojo/node!fs',
 	'dojo/node!mimetype',
+	'dojo/node!url',
 	'dojo/aspect',
 	'./util'
-], function (require, lang, Promise, http, path, fs, mimetype, aspect, util) {
+], function (require, lang, Promise, http, path, fs, mimetype, url, aspect, util) {
 	/* jshint node:true */
 
 	function Proxy(config) {
@@ -32,7 +33,10 @@ define([
 
 		_handler: function (request, response) {
 			if (request.method === 'GET') {
-				if (/\.js(?:$|\?)/.test(request.url)) {
+				if (/\/__resolveSuites__\?/.test(request.url)) {
+					this._resolveSuites(request, response);
+				}
+				else if (/\.js(?:$|\?)/.test(request.url)) {
 					this._handleFile(request, response, this.config.instrument);
 				}
 				else {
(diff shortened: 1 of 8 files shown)


kitsonk:
For sure, someone is going to run config with globs in the browser.  It is unclear from this PR how that will be handled.  Also, wouldn't it make sense to have a test for that behaviour as well?

jason0x43:
As I mentioned, in the browser context the suites list is sent back to Intern's proxy for resolution. So if the config has something like `suites: [ 'app/tests/**/*' ]`, the Intern proxy will resolve it to a list of modules and send the result back back to the browser.

There aren't any `Proxy` tests at all at the moment (which is where the test for the browser resolution functionality would go). I'll work on that.

kitsonk:
And what happens when `client.html` is used as the runner?

jason0x43:
Ah, I see. Revising...

csnover:
Did you see the already open PR for this at https://github.com/theintern/intern/pull/402 ?

jason0x43:
Many moons ago. The approach is similar to what I did, although I moved the glob resolving code to util so it could be shared with Proxy.

I have something for client.html, I'm just trying to get it to report errors cleanly and similarly in the browser when using a regular static server vs the Intern proxy.

jason0x43:
This has been revised to handle the `client.html` case. Basically, it still always tries to send suites back to the proxy for resolution. If the proxy there and no suites use globs, everything is fine. If a suite mid uses globs and the proxy isn't present, the test run will fail.

The updated commit also includes some basic unit tests for `lib/Proxy.js`. It's not full coverage yet, but it exercises the functionality used for suite resolution and provides a basis for further tests.

jason0x43:
Merged in 86193c293977ad70c74b0c0877fe2028c8d6a1c2

Issue #525, opened 2015-11-05
Browser stack overflow while serializing Suite object in WebDriver reporter


croes:
When trying to upgrade from Intern 2.2.2 to Intern 3.0.6, I've encountered the following issue.
All tests seems to run fine, but at the end of the run, the run crashes:

The following errors are reported:

```
Reporter error!
InternalError: too much recursion
  at Suite.prototype.toJSON  <__intern/lib/Suite.js:417:5>
  at sendRequest  <__intern/lib/reporters/WebDriver.js:146:12>
  at WebDriver.prototype._send  <__intern/lib/reporters/WebDriver.js:176:5>
  at WebDriver.prototype._sendEvent  <__intern/lib/reporters/WebDriver.js:130:11>
  at WebDriver.prototype.suiteEnd  <__intern/lib/reporters/WebDriver.js:54:11>
  at ReporterManager.prototype.emit/<  <__intern/lib/ReporterManager.js:250:20>
  at ReporterManager.prototype.emit  <__intern/lib/ReporterManager.js:232:23>
  at report  <__intern/lib/Suite.js:199:13>
  at end  <__intern/lib/Suite.js:193:12>
  at Promise/this.then/</<  <__intern/node_modules/dojo/Promise.js:156:41>
  at runCallbacks  <__intern/node_modules/dojo/Promise.js:19:13>
  at settle/<  <__intern/node_modules/dojo/Promise.js:103:21>
  at run  <__intern/node_modules/dojo/Promise.js:51:33>
  at CallbackQueue</CallbackQueue.prototype.drain  <__intern/node_modules/dojo/CallbackQueue.js:39:21>
  at nextTick</observer<  <__intern/node_modules/dojo/nextTick.js:52:21>

Error: One or more suite errors occurred during testing
    at http://localhost:8072/__intern/lib/executors/Executor.js:301:13
    at http://localhost:8072/__intern/node_modules/dojo/Promise.js:156:41
    at runCallbacks (http://localhost:8072/__intern/node_modules/dojo/Promise.js:19:22)
    at http://localhost:8072/__intern/node_modules/dojo/Promise.js:103:21
    at run (http://localhost:8072/__intern/node_modules/dojo/Promise.js:51:33)
    at CallbackQueue.drain (http://localhost:8072/__intern/node_modules/dojo/CallbackQueue.js:39:35)
    at MutationObserver.<anonymous> (http://localhost:8072/__intern/node_modules/dojo/nextTick.js:52:27)
```

[...]

PR #561, opened 2015-12-29
Improve suite error handling in WebDriver reporter

croes:
Fix proposal for issues described in #525.
- Add suiteError handler to WebDriver reporter.
- Detect fatal error messages in `Proxy.js`. Test execution broke when a fatal error message arrived, as there is an assert `message.sequence <= session.lastSequence`, which is not true for fatal error messages  (sequence number is 0 for fatal error messages).
- Do not set `error.relatedTest` if the test is a `Suite`. This caused infinite loops during JSON serialization (`error.relatedTest` pointed to the `Suite` itself, because that is where the error occured).
- Do not throw fatal errors if a suite error occured.
- Differentiate between suite errors and fatal errors in `Runner`.

Please review thoroughly.

### DIFF
--- a/lib/Proxy.js
+++ b/lib/Proxy.js
@@ -177,7 +177,7 @@ define([
 
 		_publishInSequence: function (message) {
 			var session = this._getSession(message.sessionId);
-			if (!isFatalErrorFromPreExecutor(message) && message.sequence <= session.lastSequence) {
+			if (message.sequence <= session.lastSequence && !isFatalErrorFromPreExecutor(message)) {
 				throw new Error('Repeated sequence for session ' + message.sessionId + ': ' + session.lastSequence +
 					' last ' + message.sequence + ' cur');
 			}
--- a/lib/Proxy.js
+++ b/lib/Proxy.js
@@ -11,6 +11,10 @@ define([
 ], function (require, lang, Promise, http, path, fs, mimetype, aspect, util) {
 	/* jshint node:true */
 
+	function isFatalErrorFromPreExecutor(message) {
+		return message.payload[0] === 'fatalError' && message.sequence === 0;
+	}
+
 	function Proxy(config) {
 		this.config = config;
 	}
@@ -173,8 +177,7 @@ define([
 
 		_publishInSequence: function (message) {
 			var session = this._getSession(message.sessionId);
-
-			if (message.sequence <= session.lastSequence) {
+			if (!isFatalErrorFromPreExecutor(message) && message.sequence <= session.lastSequence) {
 				throw new Error('Repeated sequence for session ' + message.sessionId + ': ' + session.lastSequence +
 					' last ' + message.sequence + ' cur');
 			}
--- a/lib/executors/Executor.js
+++ b/lib/executors/Executor.js
@@ -297,10 +297,6 @@ define([
 				})
 				.finally(lang.bind(self, '_afterRun'))
 				.then(function () {
-					if (self._hasSuiteErrors) {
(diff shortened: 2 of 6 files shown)


lib/Proxy.js
```diff
@@ -173,8 +177,7 @@ define([
 
 		_publishInSequence: function (message) {
 			var session = this._getSession(message.sessionId);
-
-			if (message.sequence <= session.lastSequence) {
+			if (!isFatalErrorFromPreExecutor(message) && message.sequence <= session.lastSequence) {
```
croes: An alternative would be to generate valid sequence numbers for fatal error messages, but I didn't quite figure out how to implement this.

lib/executors/Executor.js
```diff
@@ -297,10 +297,6 @@ define([
 				})
 				.finally(lang.bind(self, '_afterRun'))
 				.then(function () {
-					if (self._hasSuiteErrors) {
-						throw new Error('One or more suite errors occurred during testing');
```
croes: Should fatal errors be thrown if a suite error has occured? I would assume not, as suite errors have their own handlers. Second opinions?

lib/reporters/Runner.js
```diff
@@ -16,7 +16,8 @@ define([
 		config = config || {};
 
 		this.sessions = {};
-		this.hasErrors = false;
+		this.hasFatalErrors = false;
+		this.hasSuiteErrors = false;
```
croes: Reporting suite errors as fatal errors was confusing. I changed it so that suite errors are actually reported as suite errors and fatal errors as fatal errors.

kitsonk:
I have some unit tests for the WebDriver reporter in #451.  If this PR makes sense, we should find a way to integrate/port over those unit tests, even if we don't like the functionality that is in #451 (of course merging #451 and rebasing might also make sense).

lib/Proxy.js
csnover: Is this to prevent out-of-sequence errors if the client restarts and then sends a fatal error, or…?

lib/executors/Executor.js
csnover: The overall test run should fail if there is a suite error which is what this was doing. I don’t think it really matters how it comes out at the end as long as it’s a non-zero exit code for the run.

lib/Proxy.js
croes: Yes, exactly. If you have any suggestions on how to implement this in a better way, please let me know :)

lib/executors/Executor.js
croes: True, re-adding this so it exits with a non-zero exit code. I also had to fix serialization of fatal errors.

dylans:
Sorry for the significant delay in reviewing this @croes . @jason0x43 please review when you have a chance.

Also, @croes , can you verify if you have a [CLA](https://contribute.jquery.org/CLA/) on file?

jason0x43:
Th WebDriver reporter updates have been merged in 1d7a5d5. The other issues were handled a bit differently (but to similar effect):

* Rather than have Proxy treat fatal errors specially, the WebDriver reporter and PreExecutor now share communication code so that the sequence number is always consistent.
* `error.relatedTest` is still always set, because conceptually a nested suite is a "test" in its parent. However, Suite now checks to see if the related test is the Suite itself during serialization.
* I left out the fatal vs suite error differentiation because it required additional changes to be complete, and while it did provide arguably more specific information about errors in some cases, it introduced new ambiguities in others.

kitsonk:
Coolness! 👍

PR #562, opened 2015-12-31
Updates to Geezer Dependencies

kitsonk:
SauceLabs tunnel appears to "broken" on Node 0.10 and Geezer appears to be broken on Node 0.12+.  Allow infinite recursion to occur.  This patch updates the dependencies to DigDug and LeadFoot which appear to introduce no regressions (although the self tests can't run, because they use Geezer 2.2.2 which is broken on SauceLabs :confounded:).

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,8 +14,8 @@
 		"istanbul": "0.2.16",
 		"source-map": "0.1.33",
 		"dojo": "1.9.4",
-		"leadfoot": "1.2.1",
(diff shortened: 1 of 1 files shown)


jason0x43:
I believe the problem was due to an issue we fixed in Dig Dug 1.3.0 where Intern running under Node 0.12 wasn't properly waiting for the SauceConnect tunnel to start.

Issue #408, opened 2015-06-09
Gulp task


stimms:
I see there is a grunt task for running tests but I didn't see an equivalent gulp task or anything in the backlog here to create one. I sure would find one useful as would, I'm sure, many other people.

phated:
There's no real reason to create a gulp task, as the grunt task just spawns a child process.  Gulp lets you do that with standard node libraries

sholladay:
+1

I do weird things like [compute the user's name](https://github.com/sindresorhus/fullname) and overwrite the config file on disk just to shove it into `capabilities.name`, which helps a ton on my development team to keep track of issues on BrowserStack and understand where they are coming from. I am hoping the [vinyl-fs](https://github.com/wearefractal/vinyl-fs) approach in Gulp could be leveraged to speed this up by avoiding multiple file reads.

Even if this cannot work, it would help to have a clean, configurable API maintained and supported by Intern for how it wants to manage the child process, and which is dead simple to drop into a build system.

phated:
@sholladay creating a plugin for this goes against the gulp plugin guidelines, using vinyl-fs as a globbing mechanism is an anti-pattern.

stimms:
I made a first pass at implementing this at https://github.com/stimms/gulp-intern but it is very rough

sholladay:
@phated almost every single competitor to Intern has a Gulp task.
 - [QUnit](https://www.npmjs.com/package/gulp-qunit)
 - [Mocha](https://www.npmjs.com/package/gulp-mocha)
 - [Jasmine](https://www.npmjs.com/package/gulp-jasmine)
 - [Buster](https://www.npmjs.com/package/gulp-busterjs)
 - [Karma](https://www.npmjs.com/package/gulp-karma)
 - [Nightwatch](https://www.npmjs.com/package/gulp-nightwatch)

In addition, most of these are official plugins, sanctioned and listed at [gulpjs.com/plugins](http://gulpjs.com/plugins/).

On top of that, they get plenty of use and are a critical part of any build process.

So why not Intern? Why does Intern, specifically, not belong in the Gulp ecosystem? I have seen people use this line before, elsewhere, without giving alternatives that are better (in my mind, the DIY approach is not better). Even if there is a decent explanation, it sounds like an issue that should be addressed within Intern itself.

So where is the disconnect here? We want a test system and we want a build system, are we simply fools for wanting a bridge between the two that doesn't have to be maintained in-house, per-project? Which part of this use case is illegitimate and why?

[...]

PR #570, opened 2016-01-08
Add gulp usage information

jason0x43:
This PR adds some basic information about using Intern with Gulp.

References #408


jason0x43:
Merged in 0e4eccf

Issue #386, opened 2015-04-26
Support skipping all tests in a suite if skip() is called in setup()


Arnavion:
- Add a skip() method to Suite
- If Suite.skip() is called inside Suite.setup(), pre-emptively mark all tests in the suite as skipped without running them.

This would reduce the boilerplate needed if the entire suite needs to be disabled based on some runtime decision:

```javascript
tdd.suite("Check foo", function () {
	var fooSupported = false;

	tdd.before(function () {
		fooSupported = ...;
	});

	tdd.test("Test1", function () {
		if (!fooSupported) {
			this.skip("foo not supported in this environment.");
		}

		...
	});

	tdd.test("Test2", function () {
		if (!fooSupported) {
			this.skip("foo not supported in this environment.");
		}

		...
	});
});
```

into

```javascript
tdd.suite("Check foo", function () {
	tdd.before(function () {
		var fooSupported = ...;
		if (!fooSupported) {
			this.skip("foo not supported in this environment.");
		}
	});

	tdd.test("Test1", function () {
		...
	});

	tdd.test("Test2", function () {
		...
	});
});
```

sholladay:
+1, yes please.

Having to work around this is a common pattern for me.

jason0x43:
Fixed in 4e42a7a3e64580c86c2e5175e15793d50f43da82

PR #571, opened 2016-01-11
Allow suites to be skipped

jason0x43:
Add suite skip support

- Add a `Suite#skip` method to skip a suite. This method may be called from within a suite lifecycle method or from within a test (as `this.parent.skip()`).
- Support a new `Suite#skipped` property on suites. This property behaves similarly to `Test#skipped`. Setting it to a non-null value will cause the remainder of a suite (all tests and nested suites) to be skipped.

References #386

### DIFF
--- a/lib/ClientSuite.js
+++ b/lib/ClientSuite.js
@@ -62,6 +62,8 @@ define([
 
 					case 'suiteEnd':
 						suite = arguments[1];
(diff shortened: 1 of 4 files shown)


jason0x43:
Merged in 4e42a7a3e64580c86c2e5175e15793d50f43da82

PR #574, opened 2016-01-15
Allow cwd to be specified in grunt task

jason0x43:
This PR lets the grunt task use a `cwd` option.


jason0x43:
Merged in 9092e46

PR #575, opened 2016-01-15
Will fix #296 by adding skipped test information to summary...

### LINKED ISSUE #296: show the number of skipped tests in the html reporter summary
it would be useful to have the html reporter summary give an indication of how many tests were skipped.

jacobroufa:
… and greying out skipped test rows. The only potential modification to this that I could think of would be to update the icon to be more representative of "Skipped" but I felt a check mark would suffice.

screen shot 2016-01-15 at 1 15 38 pm

screen shot 2016-01-15 at 1 15 49 pm


sholladay:
Another approach would be to replicate the purple that shows up on the command line, for consistency.

jason0x43:
Two issues.

1. Tests skipped via `grep` and `Suite#skip` aren't showing up because the executor doesn't emit `testEnd` for those, just `testSkip`, so the reporter will need to listen for skipped tests.
2. Let's make skipped rows faded (maybe just decrease the opacity of the text in the row) rather than making the entire background stand out so much. Also, use a new icon, maybe pe-7s-less (circle with a line in it; this will require generating a new icon font in `variables.styl`), and make it purple or light gray.

dylans:
@jacobroufa want to address @jason0x43's feedback so we can land this?

jason0x43:
This has been merged

PR #577, opened 2016-01-25
Add check for if excludeInstrumentation is a string

JSMike:
Currently intern expects excludeInstrumentation to be a regular expression or a boolean true. When using intern as a grunt task and attempting to override your config file's default excludeInstrumentation value, grunt will spawn the task with the excludeInstrumentation option through the command line, converting the option to a string. This is never checked, and causes an error for /lib/util.js:760 (texcludeInstrumentation.test is not a function).

The fix that I've implemented checks the config.excludeInstrumentation variable before enableInstrumentaiton() is called in /lib/executors/Executor. If excludeInstrumentation is a string it will try to parse it (to see if it's a Boolean value), otherwise it will create a RegExp from the string. This also provides additional flexibility for intern config files, excludeInstrumentation can now either be a regular expression, or a string that will get parsed as a regular expression.

### DIFF
--- a/lib/executors/PreExecutor.js
+++ b/lib/executors/PreExecutor.js
@@ -119,6 +119,12 @@ define([
 				}
 			});
 
+            if (kwArgs.excludeInstrumentation === 'true') {
+                kwArgs.excludeInstrumentation = true;
+            } else if (typeof kwArgs.excludeInstrumentation === 'string') {
+                kwArgs.excludeInstrumentation = new RegExp(kwArgs.excludeInstrumentation);
+            }
+
 			this.getArguments = function () {
 				return kwArgs;
 			};


JSMike:
Travis got stuck on testing IE again. Closing/opening to restart Travis.

JSMike:
Travis is doing it's thing, hanging at testing IE 10/11 without any output for 10 min. I don't want to spam you guys with closing/opening, but this appears to happen frequently.

jason0x43:
I like the idea. A better place to handle this would be in `PreExecutor#getArguments`, where command line argument preprocessing takes place. Also, the code should check for `excludeInstrumentation === 'true'` rather than using `JSON.parse`, because `JSON.parse` could return many types (string, number, object, array). The only valid values for `excludeInstrumentation` are `true` (to exclude everything) or a regular expression (to exclude some things).

JSMike:
@jason0x43 Thx, I'll look there and modify the PR.

JSMike:
@jason0x43 sorry for the delay, I just got priority to work on this during our sprint. I've updated the location that you suggested, and it's currently working with my grunt watch config that spawns intern.

jason0x43:
Merged in 9410959. Thanks!

JSMike:
Thank you :)

Issue #289, opened 2014-10-15
add async support to suite object


jstrimpel:
For certain use cases it makes sense to have the convenience of `this.async` in the suite object in addition to the test object. I would like to add it as an enhancement. 

Currently the code is in https://github.com/theintern/intern/blob/master/lib/Test.js#L69. My first inclination is to put it in a module named `lib/asnyc.js` and include it `lib/Test.js` and `lib/Suite.js`. Is there another pattern, naming convention, etc. that you would prefer?

herby:
Yeah, I want be able to `this.async` in {before,after}{,Each} as well.

jason0x43:
Does anyone have an example of how this feature would be used? Suites are already run asynchronously, and `setup` and `teardown` now support async operations if a Promise is returned. Is the goal just to provide one of Intern's augmented Deferreds in the Suite lifecycle methods?

herby:
Oh, if `setup` and `teatDown` are fine by returning `Promise`, than I am ok with that.

csnover:
@jason0x43 it’s for convenience so people do not need to create their own promises to do async stuff in suite methods. it also helps to provide a more consistent api for test suites (i.e. there should be no case where `this.async` is available in one context and not in another context, even though they both do the same kind of async operation).

jason0x43:
Do they actually do the same operation? I mean, I can see value in using `this.async` to get a promise in a suite lifecycle method, but would `async`'s arguments be useful in a Suite context?

csnover:
yes, i think you would want to be able to avoid hangs caused by your suite methods never resolving too.

csnover:
i mean, the Suite ones are slightly different because they are different for each lifecycle method, whereas the Test one is directly associated with just the one test method used by the Test object. but to the end-user they are exposed the same way.

jason0x43:
So saying `this.async(1000)` in a suite's `setup` method would put a 1000 second timeout on just the setup method, while calling it in a `beforeEach` would do so for each call of `beforeEach`?

csnover:
well, 1000ms at least :)

it is just a convenience function to prevent boilerplate like this:

```js
setup() {
  return new Promise((resolve, reject) => {
    const timer = setTimeout(() => { reject(new Error("Timed out")); }, 5000);
    doTheActualWork().then(
      (value) => { clearTimeout(timer); resolve(value); },
      (error) => { clearTimeout(timer); reject(error); }
    );
  });
},
beforeEach() {
[...]

PR #581, opened 2016-01-29
Add suite async support

jason0x43:
- Suite lifecycle methods can now call `this.async` to enable async
  functionality. Calling `this.async` only affects the particular
  lifecycle method call in progress.
- The lib/Suite unit tests were reorganized, and async tests were added
  for the lifecycle methods.

References #289

### DIFF
--- a/lib/Suite.js
+++ b/lib/Suite.js
@@ -1,7 +1,7 @@
 define([
 	'dojo/Promise',
-	'dojo/lang'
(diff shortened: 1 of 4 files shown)


jason0x43:
Merged in https://github.com/theintern/intern/commit/63e09383ede6502f80153a54461d532cb54e7f69

Issue #413, opened 2015-06-12
Feature request: Fail-fast option


sholladay:
### Problem
For users of Continuous Integration / Delivery systems, the excess testing and build time caused by multiple test failures can be very undesirable, such as for a shared dev branch where speed is important and time-to-deploy is compounded and queued many times over. It can also consume precious resources from per-hour test clouds like BrowserStack and SauceLabs.

### Solution
Provide an opt-in configuration option to halt at the first test failure. A relevant reference is the [fail_fast](https://www.relishapp.com/rspec/rspec-core/v/2-0/docs/configuration/fail-fast) option for RSpec or Mocha's `--bail`.

### Suggested requirements
 - When a test fails, all further tests and suites are aborted.
 - Any tests not run due to a test failure are counted as skipped.
 - When tests are skipped due to a test failure, a warning is logged to explain them.
 - This option is off by default, to avoid confusion.

Stretch goal: make it configurable so that it can be told to only abort the rest of the current suite or only the rest of the unit tests, but not functional tests, etc.

vladikoff:
Good suggestion, I do agree that this is gonna useful for CI, especially if you retry tests with something like `travis_retry`.


<img src="http://static.comicvine.com/uploads/scale_super/5/52246/2060390-i_like_it.jpg" width=150 />

sholladay:
It can be even better. If Intern natively [supports retries](https://github.com/theintern/intern/issues/340), it doesn't have to repeat the entire lot of suites or even restart the process. In my ideal setup, CI would probably retry a test once or twice before considering it truly failed and once that had been decided, would not continue. So combining these is indeed powerful, especially if it's handled within Intern itself.

vladikoff:
@csnover in https://github.com/theintern/intern/issues/368#issuecomment-88585145 you mentioned that in Intern 3 it should be possible to pause flow, what this help this issue?

csnover:
Yes it will help, it should be much easier to bail out of running any more tests now

vladikoff:
any pointers on how to do that? I think @miketaylr wanted to hook that up

csnover:
In `lib/Suite.js`, if a test fails and fast fail is turned on then don’t run any more tests. This changed code also needs to account for sub-suites: a parent suite should not continue to run either if a child suite failed. That’s pretty much it.

airtonix:
Any way I can enable fail fast but not have my terminal spammed with all the skipped tests?

jason0x43:
[...]

Issue #340, opened 2015-01-23
Feature Request: Automatic retries for failed tests


sholladay:
### Problem
Some types of tests, particularly functional tests and any that rely upon a network, are prone to sporadic reliability problems. For users of Continuous Delivery systems, this is extremely undesirable. People get e-mailed, passwords are needed to cleanup the logs, etc.

### Solution
Some opt-in machinery to automatically retry failed tests, which will help smooth over minor timing glitches.

**Example**:
An `intern.js` configuration file could have:
````javascript
...
// Default : 1
maxTries : 3
...
````
... and that would cause any given test to be able to run up to 3 times total.

**Comment by csnover** [comment]
Yes, this is a possibility. In the interim you can just write a simple helper function to perform a retry. Something like this might work (not tested):

```js
function retry(maxRetries, test) {
  var numExecutions = 0;

  return function () {
    var self = this;
    var args = arguments;

    function retry(lastError) {
      var dfd = new Deferred();
      var promise;

      if (++numExecutions <= maxRetries) {
        var result;
        try {
          result = test.apply(self, args);
          dfd.resolve(result);
        }
        catch (error) {
          result = error;
          dfd.reject(result);
        }

        promise = dfd.promise.otherwise(retry);
      }
      else {
        dfd.reject(lastError);
        promise = dfd.promise;
      }

      return promise;
    }

    return retry();
  };
}

registerSuite({
  foo: retry(3, function () {
    // this test will be run up to 3 times
  })
});
```

sholladay:
Great! And thanks for the example snippet.

vladikoff:
@jason0x43 this can be closed I think

sholladay:
Yeah, I take it back. Tried to be level headed about it, but this is the making of a footgun.

neilk:
It's generally a bad idea to retry tests a lot. But in case anyone else really needs this (I do, for a temporary situation), here's a simple `retry` implementation that is working for me.

```
 function retry (maxTries, p) {
    var tries = 0;
    var iter = function() {
      tries++;
      return p().catch(function(error) {
        console.log(">>> Test errored on try " + tries + "/" + maxTries)
        if (tries <= maxTries) {
          return iter();
        } else {
          return Promise.reject(error);
        }
      });
    };
    return iter();
  }
```

PR #582, opened 2016-01-29
Add fail-fast and test retry functionality

jason0x43:
- Add a new config option `failFast`. Setting this to a truthy value
  will cause all subsequent tests after a failing test to be skipped.
- A new config option `retries` that specifies the number of
  times a test should be retried if it fails. This value may also be set
  for individual tests or suites within the test or a suite lifecycle function.

References #413, #340

### DIFF
--- a/lib/ClientSuite.js
+++ b/lib/ClientSuite.js
@@ -62,6 +62,8 @@ define([
 
 					case 'suiteEnd':
 						suite = arguments[1];
+						self.skipped = suite.skipped;
+
 						// The suite sent by the server is the root suite for the client-side unit tests; update the
 						// existing test objects with the new ones from the server that reflect all the test results
 						if (!suite.hasParent) {
--- a/lib/Suite.js
+++ b/lib/Suite.js
@@ -11,6 +11,10 @@ define([
 		this.reporterManager && this.reporterManager.emit('newSuite', this);
 	}
 
+	// FAIL_FAST_REASON needs to be a string so that Intern can tell when a remote has skipped unit tests so that it can
+	// skip functional tests.
+	var FAIL_FAST_REASON = 'fail fast';
+
 	Suite.prototype = {
 		constructor: Suite,
 		name: null,
@@ -22,16 +26,29 @@ define([
 		teardown: null,
 		error: null,
 		timeElapsed: null,
+		_failFast: null,
 		_grep: null,
 		_remote: null,
 		_environmentType: null,
 		_reporterManager: null,
+		_retries: null,
 
 		/**
 		 * If true, the suite will only publish its start topic after the setup callback has finished,
 		 * and will publish its end topic before the teardown callback has finished.
 		 */
 		publishAfterSetup: false,
+
+		/**
+		 * A flag used to indicate whether a test run shoudl stop after a failed test.
+		 */
+		get failFast() {
+			return this._failFast || (this.parent && this.parent.failFast);
+		},
+
+		set failFast(value) {
+			this._failFast = value;
+		},
 
 		/**
 		 * A regular expression used to filter, by test ID, which tests are run.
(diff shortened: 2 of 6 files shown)


vladikoff:
I'm excited for this feature, thanks for working on this!

sholladay:
Some bikeshedding on the names:
 - I vote for `tries`, with a default of `1`. Seems easier to read / faster to type / more intuitive to me.
 - Some other test systems, namely Mocha, say `bail` instead of `failFast`. I don't feel like either name is particularly descriptive, but would probably choose whichever is more conventional / familiar to people.

jason0x43:
I like `bail` vs. `failFast`, and `bail` does seem to be the common term.

I still prefer `retries` to `tries`. In general, all tests should run at least once unless explicitly skipped, and skipping should be done with `Suite#skip`, `Test#skip`, or `grep`, which provide a reason for skipping.

sholladay:
Yeah explicitness is good, true that. Hooray for `bail`. I was starting to realize I would routinely mix up the order, "fail fast" vs "fast fail".

spalger:
:+1: excited to see this merged

jason0x43:
Soon...

sholladay:
To clarify, the expected behavior is that `retries` takes precedence over `bail`, is that correct? In other words, `bail` typically means to exit upon the first failure, but that won't happen unless the test remains in a failed state _after_ its allotted number of `retries`. That is something I could see people getting confused about, the order of precedence. I hadn't given that much thought before. Is it intuitive?

Also, would `retries` affect a suite that errored in one of its lifecycle methods? And for failed tests that are being retried, will their `beforeEach` methods get called?

The more I think about `retries`, the more wary I am of it. Since I originally filed #340, I have figured out techniques to make functional tests more resilient to transient timing glitches. So the value of it vs its maintenance cost and pitfalls seems blurry to me now. I was always concerned that it would mask genuine problems in a test suite, but it seemed worth taming the WebDriver beast. I think it only makes sense if a lot of tests are unnecessarily messy without it. What's the reality like for people?

jason0x43:
1) Yes, `retries` takes precedence over `bail`, in that the executor won't bail until a test has failed all of its `retries`. As far as Suite is concerned, a test hasn't really failed if it still has retries available, so bailing won't kick in.
2) `retries` only comes into play for errors in the test functions themselves, not for errors from suite lifecycle methods. So an error in `beforeEach` would not result in a test being retried. This seems reasonable to me since the lifecycle methods aren't part of any particular test (but the case could be made that they should invoke retries).
3) When a test is retried, `beforeEach` does get called again. Each try of a test is a complete run of that test (`beforeEach`, test, `afterEach`).

I agree, `retries` are a tricky feature. Ideally, tests should either always pass or always fail, and making it easy to retry tests could also make it easier to write lower quality tests. On the other hand, WebDriver tests can be unstable due to both browser and network related timing issues, and I could see `retries` being a reasonable solution.

When writing functional tests myself, I've generally attempted to make tests more robust, either by adjusting timings in individual tests or by implementing more robust operations in Command helpers that I can use in multiple tests. Something similar may be a preferable solution here; Intern or Leadfoot could include a wrapper function that would add retry functionality at the individual test or Command chain level. The main issue for that type of solution is that the user has to be more aware of what will be retried -- the wrapper function would be handling retries, so the suite lifecycle methods would not, by default, be re-run.

spalger:
We implemented a method similar to the one you described @jason0x43 and it has proved very useful when dealing with WebDriver.

Ours is called [`tryForTime`](https://github.com/elastic/kibana/blob/master/test/support/pages/common.js#L184-L208), it accepts a function and timeout and retries the function until its promise resolves or the timeout is reached. It behaves a lot like the `find` + `findTimeout` combo.

jason0x43:
Hmmm...I think I may split retry functionality out into a separate PR for further analysis. On the one hand, making it globally available could lead to abuse (using retries to mask poor tests). On the other hand, one of Intern's goals is to make test writing easier, and I'd prefer not to end up with a more complex solution solely to keep people from shooting themselves in the foot.

sholladay:
That makes perfect sense to me. `bail` doesn't seem like much of a rabbit hole, beyond figuring out how to communicate the situation to the user clearly.

jason0x43:
Support for `bail` was merged in https://github.com/theintern/intern/commit/a6a732ef4864f23f68aee3bbb4ebb27c888f0992.

kailo777:
It there are reason why the suppport for retry was not merged?

jason0x43:
Because of the reasons stated above (i.e., at the time it wasn't clear that a generic retry feature was really a good thing to add). If it seems worthwhile, open a new issue and we can take a look again.

PR #585, opened 2016-02-11
[util] fixes #584 - Consistent handling for URLs during path normalization

### LINKED ISSUE #584: Path normalization breaks URL
Path normalization code in lib/util.js transforms `http://domain.tld/path/to/file` into `http:/domain.tld/path/to/file` , which breaks all subsequent resolution of relative paths, specially when invoking `switchLoader()` .

olemis:
Detects whether input path is a URL and preserves <scheme>:// prefix .

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -690,7 +690,8 @@ define([
 			}
 
 			var parts = path.replace(/\\/g, '/').split('/');
-			var result = [];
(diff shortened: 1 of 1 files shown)


jason0x43:
I made an alternative fix in 75c1472d6d1e3fcf7b1ed7f92deb14ea8b631a44.

PR #590, opened 2016-03-07
Update intern to use latest dojo-loader/loader and dojo-core plugins

tomdye:
Addresses: https://github.com/dojo/loader/issues/49
- Swaps out usage of `dojo/loader` for the new `dojo-loader/loader` - Uses latest `dojo-core/text` plugin
- Uses latest `dojo-core/has` plugin
- Adds `selftest.local.sh` for running tests against local `chromedriver` - Sets loader to `dojo-loader/loader` in `selftest.intern.js` - Code needs to use `has.default()` until intern upgraded to TS


jason0x43:
A couple of thoughts...

- fixdeps should be updated to link dojo-loader into `browser_modules`, and that should be referenced in `client.html`
- If we're using the new loader as the default loader, we shouldn't need to set `loaders` in the test config
- I'm not sure we need to include selftest.intern.local, but if we did, it should just import selftest.intern and modify the couple of fields that need to be updated rather than duplicating the config

jason0x43:
I had a bit more time to play with this over the weekend. So, to just use `dojo-loader` as Intern's loader, you don't need to specify the loader in Intern's config. Doing so will use the loader in self-tests with the testing version of Intern, but isn't necessary for using the loader with Intern itself when testing some other project. Also, when using `dojo-loader`, but using the existing dojo2 for everything else, you don't need to import core for `dojo/text`, because the `dojo/text` plugin should still work.

tests/selftest.intern.js
```diff
@@ -16,6 +16,11 @@ define({
 
 	maxConcurrency: 2,
 	tunnel: 'SauceLabsTunnel',
+
```
Tomdye: I needed to add the `loaders` config here to make the tests run with the `dojo-loader` otherwise I couldn't prove that it was working

Tomdye:
@jason0x43 points addressed. Have maintained the setting of custom loader in the selfTest.local because otherwise the older version of `intern` does not use the `dojo-loader` when running selfTest, at least this way it will use the new `dojo-loader` locally.

Tomdye:
`dojo-core/has!` plugin now used instead of `dojo/has!`

kitsonk:
I was thinking about this, the biggest challenge is trying to use the `dojo-core` modules in their "SymstemJS ES6" format where there is a default export which gets mapped to `[module_factory_return].default` and tagged with some meta data.

We could of course just change those modules which Intern uses back into "standard" modules, where there is `export = foo;` instead of `export default foo`, when importing to untranspiled AMD, it would work in the old way.

It looks based on this PR, it would only need to be `dojo-core/has`.  It would make a few changes into `dojo-core` to revert imports of `import has from './has'` to `import * as has from './has'` but that isn't the worst thing in the world.

psingh4:
what is the plan to release/merge this? or is there a workaround/interim solution available for using es6 modules with intern to run in the node environment?

dylans:
@psingh4 this is probably already handed with https://github.com/theintern/intern/issues/674 which landed in master last week. There's still a fair amount of work to be done, but Intern master now leverages the real Dojo 2 and supports writing tests using ESM.

jason0x43:
As @dylans noted, this functionality will be a part of Intern 4 (or at least, using Babel with Intern 4 will be a bit more straightforward). In the meantime, you should be able to use an intermediate Node-suite loading module, along with babel-register, to load ES6 suites in Node, something like this:

```ts
// intern.js
suites: [ 'tests/nodeSuites' ]
```

```ts
// tests/nodeSuites.js
define([
    'intern!object',
    'dojo/node!babel-register'
], function (registerSuite) {
    require('dojo/node!tests/suite1.js')(registerSuite);
    require('dojo/node!tests/suite2.js')(registerSuite);
});
```

```ts
// tests/suite1.js
module.exports = registerSuite => {
    registerSuite({
        name: 'suite1',
        // ...
    });
};
```

Note that the code above hasn't been tested. Also note that this will, of course, only work in a Node environment.

jason0x43:
This has, I think, been overcome by events

Issue #606, opened 2016-04-05
Pretty reporter shouldn't show coverage table when coverage is disabled


jason0x43:
The table header and footer are still printed when `excludeInstrumentation` is set to true, which completely disables coverage.

jason0x43:
Fixed in 12d0acc

PR #609, opened 2016-04-07
Hide coverage table when list of files is empty

LingSpb:
This PR fixes the issue https://github.com/theintern/intern/issues/606
It hides the coverage table not only when excludeInstrumentation set to true, but in more generic cases - when the list of files is empty


jason0x43:
Merged in 12d0acc. I modified the check to be against `files().length` rather than `store.keys()` because `files` is part of the published API. Thanks!

PR #616, opened 2016-04-13
Allow passing in options for istanbul's Instrumenter through intern's configuration.

Arnavion:
The instrumentOptions property will be merged into the default options object passed to istanbul's Instrumenter constructor.

---

intern 3.1 uses istanbul 0.4.1 which uses escodegen 1.7.x. Starting from this version escodegen converts unicode string literals to utf-8 bytes more frequently, and requires an option `verbatim: "raw"` to revert to the old behavior. This breaks my testing and I've opened https://github.com/gotwarlost/istanbul/issues/591 to get their view on whether istanbul should set this by default or users of istanbul should.

Regardless of how that's decided, specifying arbitrary options for `Instrumenter` may be useful for other people for other reasons in the future, so I figured I'd make it available in any case.

```irc
<Arnavion> Does anyone have an opinion on   https://github.com/gotwarlost/istanbul/issues/591   ? i.e. should it be istanbul's responsibility to enable `verbatim: "raw"` in the options it gives to escodegen by default? Or should intern specify it in the options it gives to istanbul? Or should intern accept this option from the user?
<Arnavion> As it currently stands I'm unable to move from intern@3.0.x to 3.1.x because of this
<jason0x43> So the previous behavior was the equivalent of setting `verbatim: 'raw'`, and if the option can be set as an option during Instrumenter creation, I think it's reasonable for Intern to do that (and, in fact, is desirable since it would restore the previous default behavior).
<jason0x43> *So if the previous behavior…
<jason0x43> Although Istanbul doing it would make more sense if it was a breaking change from previous versions.
<Arnavion> If my case is a fringe case I'm more than happy to pass in the option myself to intern and have it propagated to istanbul
<Arnavion> (Regardless of this particular setting, intern might want to allow that in any case)
```

### DIFF
--- a/lib/Proxy.js
+++ b/lib/Proxy.js
@@ -153,7 +153,6 @@ define([
 						data = self._codeCache[wholePath] = util.instrument(
 							data.toString('utf-8'),
 							wholePath,
-							self.config.coverageVariable,
 							self.config.instrumenterOptions
 						);
 						send(contentType, data);
--- a/lib/Proxy.js
+++ b/lib/Proxy.js
@@ -153,7 +153,8 @@ define([
 						data = self._codeCache[wholePath] = util.instrument(
 							data.toString('utf-8'),
 							wholePath,
-							self.config.coverageVariable
+							self.config.coverageVariable,
+							self.config.instrumenterOptions
 						);
 						send(contentType, data);
 					});
--- a/lib/executors/Executor.js
+++ b/lib/executors/Executor.js
@@ -30,7 +30,9 @@ define([
 		 * @type {Object}
 		 */
 		config: {
-			coverageVariable: '__internCoverage',
+			instrumenterOptions: {
+				coverageVariable: '__internCoverage'
+			},
 			defaultTimeout: 30000,
 			reporters: []
 		},
@@ -86,7 +88,6 @@ define([
 					return self.enableInstrumentation(
 						config.basePath,
 						config.excludeInstrumentation,
-						config.coverageVariable,
 						config.instrumenterOptions
 					);
 				}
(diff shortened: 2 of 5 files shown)


jason0x43:
Any chance you could you update this so that the internal APIs only deal with `instrumenterOptions` (instead of `instrumenterOptions` and `coverageVariable`)? We'll need to keep the `coverageVariable` config property separate for backwards compatibility, but it'd be cleaner internally to only have to pass around one options object instead of options and one separate property.

Arnavion:
Can you clarify where the separation between public and internal API is? Is `Executor` public or internal? Should I normalize the config in `PreExecutor.getConfig` or somewhere later down the pipeline?

Arnavion:
- `tests/unit/lib/util.js` apparently [calls `util.instrument`](https://github.com/theintern/intern/blob/1369b356ba20218b3b4a1dabbdbf4b9305f65e7b/tests/unit/lib/util.js#L130) without a value for the `coverageVariable`, leading to it being coerced to the string `"undefined"` when used to index the `instrumenters` object. Deliberate? For now I've re-implemented that behavior by allowing `instrumentOptions` to be undefined and initializing it to an empty object.

- I don't understand how my change caused the runner self-test failure, and I don't have a Saucelabs account to run the test myself.

jason0x43:
The external API as far as coverage variable is concerned is just the config property. So, basically, Intern should still understand `config.coverageVariable`, and `config.coverageVariable` should still be available for anyone that looks, but internally only `instrumenterOptions` should be passed around. Normalizing in `getConfig` is fine.

The unit test calling instrument without coverageVariable is fine in that case. That unit test isn't concerned with coverage, so the value of the coverage variable isn't relevant, just that instrumentation be performed.

Hmmm...I believe the current CI test failure is due to a commit I merged in that modified the Pretty reporter.

Arnavion:
Should I rebase the PR on top of your commit then?

Arnavion:
New error after rebasing:

```
FAIL: internet explorer 9.0 on Windows 7 - unit tests - intern/lib/interfaces/qunit - events - testDone
CancelError: Timeout reached on internet explorer 9.0 on Windows 7 - unit tests - intern/lib/interfaces/qunit - events - testDone
No stack or location
```

jason0x43:
I did a little bit of cleanup and added a simple unit test for `util#instrument`. Merged in 2c0e0a4.

Arnavion:
Thanks!

PR #620, opened 2016-04-15
Add new 'rawIO' option to disable stderr/stdio piping

MartinDoyleUK:
By default, the Node.JS stdio is piped through to a function to handle it—`logOutput()`—however this causes a problem with [my custom reporter](https://github.com/Alfresco/Aikau/blob/master/aikau/src/test/resources/reporters/ConcurrentReporter.js). Specifically, not all `console.log()` outputs are emitted prior to any stderr writes. What this means is that, when [outputting a large set of results to the console](https://github.com/Alfresco/Aikau/blob/master/aikau/src/test/resources/reporters/ConcurrentReporter.js#L886), and if any errors occurred in any suites, then the Executor will throw an error just as the results are starting to be output, swallowing the remaining results. For an hour-long test run with maybe a hundred or more lines of detailed test failures/errors, this is disastrous. By directly passing console output through to the parent process, my custom reporter is able to complete its output before the `self._hasSuiteErrors` test in `Executor.js` throws an error to the console, breaking the stdout output.

As further confirmation of this issue, if I have a `console.error()` statement after all of my `console.log()` statements, then that statement will output immediately prior to this Executor error, even though the majority of the preceding `console.log()` statements never make it to the console.

The fix is to permit an extra option to be passed in—`rawIO`—which allows the parent process to directly receive these statements, rather than having them piped through an event mechanism which seems to inaccurately flush its queues (this is an educated guess, but by no means prescriptive of the issue). Because this is a new option, it cannot affect any existing usage of Intern by anyone. This absolutely does fix my test case using my custom reporter.

_Full disclosure: SIGINT behaviour seems to be to kill the grunt task as well, so using `rawIO` will have that caveat until this is resolved (I can't find a quick solution at the moment)._

### DIFF
--- a/tasks/intern.js
+++ b/tasks/intern.js
@@ -113,7 +113,8 @@ module.exports = function (grunt) {
 			args: args,
 			opts: {
 				cwd: opts.cwd || process.cwd(),
-				env: env
+				env: env,
+				stdio: opts.rawIO ? 'inherit' : 'pipe'
 			}
 		}, function (error) {
 			// The error object from grunt.util.spawn contains information
@@ -121,7 +122,9 @@ module.exports = function (grunt) {
 			done(error ? new Error('Test failure; check output above for details.') : null);
 		});
 
-		child.stdout.on('data', readOutput);
-		child.stderr.on('data', readOutput);
+		if (!opts.rawIO) {   
+			child.stdout.on('data', readOutput);
+			child.stderr.on('data', readOutput);
+		}
 	});
 };


MartinDoyleUK:
Am sure that the CI tests passed with my previous (improperly signed) Pull Request which had the same changes? Is the CI result definitely correct?

MartinDoyleUK:
Also, this might be a related Node.JS bug? https://github.com/nodejs/node-v0.x-archive/issues/8329

MartinDoyleUK:
Closing and re-opening to force CI re-run

MartinDoyleUK:
@jason0x43 Might you be able to review and possibly merge this, please?

jason0x43:
Since this maps directly to `stdio`, would it be beneficial to just have the Grunt task understand a `stdio` option rather than adding a new `rawIO` boolean flag? This would be somewhat more flexible, since it would allow the grunt task to connect Intern's I/O channels to arbitrary streams or file descriptors as well as inheriting the parent's descriptors.

MartinDoyleUK:
The only problem with that is working out how to handle the conditional piping of the stdout/stderr from the child-process. Do you have any suggestions?

jason0x43:
Hmmm...what do you mean?

MartinDoyleUK:
There are just two changes in the PR. One is this rawIO option being added to the child process, but the other is to conditionally enable piping of output from the child process. I'm not entirely sure how to handle the latter, if we just let them pass through the raw `stdio` option? Should we analyse the contents of the option they pass through? That's slightly complicated by [the different ways in which you can pass through the options ...](https://nodejs.org/api/child_process.html#child_process_options_stdio)

jason0x43:
The default behavior when creating a child process is for its output to be piped to the parent, which is what the task is doing now and why it needs the stdout and stderr readers. The new `rawIO` option is just a convenience for setting stdio to the string 'inherit'. If the user passes an array, then yes, I think we can inspect it and use the current behavior for the first two entries (hook up a reader if the value is 'pipe' or null/undefined). Otherwise, assume the user knows what they're doing.

MartinDoyleUK:
Roger that. Will try and make those adjustments soon and push an update.

MartinDoyleUK:
Sorry. Been moved to a new project at work, and not had time to look at this again. I'll try to get to it when I'm able.

jason0x43:
Closing this due to lack of activity, but we can re-open it if someone wants to finish it up.

PR #625, opened 2016-04-28
Ignore code cache after modification

### LINKED ISSUE #624: Proxy should invalidate code cache if instrumented file changes
I'm using Intern to test some browser code in a [Dojo 2 package](https://github.com/dojo/routing/pull/8). I use `grunt intern:proxy` to start the proxy server, then open <http://0.0.0.0:9000/node_modules/intern/client.html?config=_build/tests/intern> in my browser. My source files are configured to be instrumented.

Due to [this line](https://github.com/theintern/intern/blob/2c0e0a499c548808b9bd177b5dc590bc22f8725b/lib/Proxy.js#L136) once my source file has been instrumented, the proxy will keep serving that version, even if I make changes.

Non-instrumented files are always [read from disk](https://github.com/theintern/intern/blob/2c0e0a499c548808b9bd177b5dc590bc22f8725b/lib/Proxy.js#L179).

I'd expect changes to be reflected when I reload the test runner. One approach would be to always load the file, but hash the filename and the content and use that to see whether the cache is stale. That way the code isn't instrumented unnecessarily, and changes are reflected. This doesn't seem worse in performance than how non-instrumented files are handled.

Caching in [nyc](https://github.com/bcoe/nyc) is implemented in a similar fashion.

Happy to do a PR.

novemberborn:
The code cache in the proxy contains instrumented code. Store the mtime of the original file along with the instrumentation data. When serving the file, if the mtime has changed ignore the cached data and instrument the code again.

Fixes #624.

I couldn't quite see whether this logic has test coverage, or indeed run the test suite locally. The change detection worked with my local project. Let me know if you want me to add tests, and where.

### DIFF
--- a/lib/Proxy.js
+++ b/lib/Proxy.js
@@ -132,53 +132,59 @@ define([
 			}
 
 			var contentType = mimetype.lookup(path.basename(wholePath)) || 'application/octet-stream';
-			if (instrument) {
-				if (self._codeCache[wholePath]) {
-					send(contentType, self._codeCache[wholePath]);
+			fs.stat(wholePath, function (error, status) {
+				// The proxy server was stopped in the middle of the file stat
+				if (!self.server) {
+					return;
+				}
+
+				if (error) {
+					self._send404(response);
+					return;
+				}
+
+				if (instrument) {
+					var mtime = status.mtime.getTime();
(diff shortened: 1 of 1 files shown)


jason0x43:
Merged in https://github.com/theintern/intern/commit/f375092837a69b8a4bbb156e7ec985094cc44cc3. Thanks!

Issue #629, opened 2016-05-11
TeamCity/Pretty: Wrong numFailedTests when a suite has error at `before`


LingSpb:
When a suiteError appears in `before`, currently that suite is considered as 1 FAIL, no matter how many tests are in it. So the output looks strange, like

> Total: [×-   ] 1/4
> Passed: 0  Failed: 1  Skipped: 0

Example:

```
bdd.describe("Suite1", function () {
bdd.before(function () {
throw Error("Suite Error before");
});

bdd.it("Test 1", function () {

});
bdd.it("Test 2", function () {

});

bdd.describe("Suite 1 child", function () {
bdd.it("Test 1 1", function () {

});
bdd.it("Test 1 2", function () {

});
});
});
```

LingSpb:
This issue is quite obvious, I wonder why no one does anything, like, at least marking it as a bug or saying that we will not support?

dylans:
@LingSpb we just have a backlog we're working through, and we hope to get to it soon. We appreciate your patience.

LingSpb:
Thanks for the response. I just wanted to know if we consider it as a bug, so that I will know whether I should spend sometime to investigate this issue on TeamCity :)

jason0x43:
This isn't really a bug. When a suite fails, Intern can't tell how many tests would end up failing because, for example, some might still have passed. Rather than guessing, it just indicates the suite failure and says nothing about the tests, because they never ran.

At the time this behavior was developed we didn't have the concept of explicitly skipped tests, so probably the most correct option would be to mark all the tests in the suite as skipped with a reason of 'suite failure' or something. We also probably shouldn't record the suite failure as a failed test, although some way of indicating a suite failure in the summary would be beneficial.

LingSpb:
I'm fine with any of: 
* (a) skip all the tests in the suite and mark them as skipped
* (b) indicate the suite as failure
* or (c) keep as it is now - consider all the tests as `failed` when the suite fails

In my view,
* with (a): Not hard to implement (I hope), but how about the case the suite fails in `beforeEach`? Currently we mark the test as `failed` when there are errors in `beforeEach`, but I think to make it consistent with `before`, we should mark it as `skipped` too?
* with (b): It sounds a nice idea but not sure it will be good in practice, like how TeamCity (or other CI) will interpret and display the test results?
* with (c): The implementation should be quite similar to (a), except, we don't need to change the code for `beforeEach` since it is marked as `failed` now --> less things to do if we choose (c)

[...]

PR #630, opened 2016-05-12
Fix wrong numFailedTests in Pretty

LingSpb:
This PR is supposed to fix the wrong numFailedTests in Pretty reporter of the issue #629 

Currently we have the same treatment for suiteError which happens in `before` and `beforeEach`. But it causes the numFailedTests to be off, because with the error in `before`, that suite will run only 1 time and 1 fail will be recorded.
By adding `relatedTest` information to the suiteError for the `beforeEach` case, we can count the numFailedTests in the reporter correctly.

### DIFF
--- a/lib/Suite.js
+++ b/lib/Suite.js
@@ -260,6 +260,9 @@ define([
 					suite.async = undefined;
 
 					if (error !== Test.SKIP) {
+						if (name === "beforeEach"){
+							error.relatedTest = args[0];
+						}
 						return reportSuiteError(error);
 					}
 				});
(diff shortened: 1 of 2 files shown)


dylans:
Thanks for the PR @LingSpb.

This looks reasonable to me though I'm not that familiar with the pretty printing code (also I'm not sure if we want to accept the changed indentation on the case statement as I think the current format is consistent with the style guide, but that's a minor point). @devpaul or @jason0x43 , please review when time permits.

dylans:
@LingSpb did you want to address the comment from @jason0x43 in https://github.com/theintern/intern/issues/629#issuecomment-221549650 ?

jason0x43:
This was handled through commits 5f88f6b and 94234d2. The first updates Suite and the reporters to not consider un-run tests as failed. The second updates Pretty's test counting logic.

PR #634, opened 2016-05-20
Added version range and version aliases

devpaul:
Adds support in Intern for version aliases described in theintern/digdug#20. This requires PR theintern/digdug#24.


jason0x43:
This was merged in ab9a24516a0e59e535d6dca2112b7c417b83ca05

PR #643, opened 2016-06-12
Implement callback api to TDD and BDD interfaces

hakatashi:
Hello. I'm a student from Japan accepted for this year's GSoC program of Intern. I'm sending this pull request as my practice for diving into Intern's codebase.

Anyway, I recently migrated my JavaScript test codes written for mocha into Intern. It was almost nicely done, but I found some difficulty when my code uses asynchronous test cases.

Though Intern's user guide states that "Intern always uses Promise," it will be nice to have callback-style syntax for TDD and BDD interfaces as "non-recommended way" to preserve compatibility with such test codes, especially for the people who already have test suites like mocha and Jasmine.

I simply wrapped the provided test function with another function which returns deferred object when the test function takes an argument. So this may be compatible with the traditional callback-style syntax at least used in mocha or Jasmine.

### DIFF
--- a/lib/interfaces/tdd.js
+++ b/lib/interfaces/tdd.js
@@ -33,6 +33,24 @@ define([
 		},
 
 		test: function (name, test) {
+			// If the test function takes an argument, wrap it with deffered object to
+			// gently run asynchronous tests against callback-style
+			// for the compatibility with other test frameworks
+			if (test.length === 1) {
+				var oldTest = test;
+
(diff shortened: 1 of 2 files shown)


jason0x43:
This PR does show a good understanding of how Intern manages test suites. However, while I can see adding adapters to allow suites written for other systems to work with Intern, I'm not sure we should add callback style support to Intern just to have it.

PR #650, opened 2016-07-14
Integrate cucumber

rhpijnacker:


rhpijnacker:
It is about half a year since this pull-request was posted. Any ideas on whether this will ever be integrated?

dylans:
@rhpijnacker I think we got to the point of debating whether Intern should be able to call out to Cucumber (which is what your PR does), or if Intern should natively support Cucumber syntax so that it could do more with that.

I think we can reasonably make a persuasive argument for both approaches, but given that this is what we have, we should consider landing it for say a 3.5 before we finish the TS refactor for Intern 4? Or perhaps its better to have this as a separate package (e.g. intern-cucumber) since this does introduce a dependency on Cucumber for all Intern users, whether the use Cucumber or not?

Either way, let's make a decision. :)

jason0x43:
I think ideally the cucumber interface should just be an interface that constructs regular Intern Suites rather than relying on Cucumber to run tests. Aside from being lighter weight, this would ensure that Cucumber tests are handled the same as other tests within Intern. The gherkin parser used by cucumber is a standalone project with no other dependencies; we should see what would be involved in using it.

rhpijnacker:
@dylans Putting this in a separate package seems to make sense. However, I could use some assistance on how to set this up, both package-structure wise (how do we get intern to know about the additional package) and location/publication wise (is just having another github repo enough?)

dylans:
@rhpijnacker I was thinking something along the lines of https://github.com/theintern/intern-visual or https://github.com/theintern/intern-a11y in terms of how they extend Intern, though admittedly I've not thought through the details enough to know if that would work well or not.

I know we do want to eventually do what @jason0x43 suggested in the previous comment, "ideally the cucumber interface should just be an interface that constructs regular Intern Suites rather than relying on Cucumber to run tests.", though I appreciate what you have now is a workable solution (and why I think it probably makes more sense as an extension in its current form if possible).

PR #652, opened 2016-07-25
Use test/suite [name] instead of [id] in TeamCity report

LingSpb:
TeamCity report wraps tests by suite. Currently in Intern we use test/suite `id` (which are made by adding the parent suite name to the current test/suite name) to display tests in the report. You can see that the reports look quite a mess with repeated information.
By using `name`, the reports will look much cleaner.
Before:
teamcity_before
After:
teamcity_after

One more thing you can notice is when the test name becomes too long (because of nested suites), TeamCity will give a wrong test number. (In my example, TeamCity says 18 tests instead of 26)

### DIFF
--- a/lib/reporters/TeamCity.js
+++ b/lib/reporters/TeamCity.js
@@ -64,16 +64,16 @@ define(['../util'], function (util) {
 		},
 
 		testStart: function (test) {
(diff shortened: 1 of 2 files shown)


LingSpb:
This PR will fix the issue #659

jason0x43:
Merged in b57de6c. Thanks!

PR #654, opened 2016-08-02
fixed broken link

sallyx:


dylans:
Thanks @sallyx . It's a link fix in the docs, so no need for a CLA.

PR #655, opened 2016-08-02
Add CoverageThreshold reporter

JSMike:
Adds reporter for verifying coverage precentage thresholds are met before continuing with CI


JSMike:
This is a followup to https://github.com/theintern/intern/pull/553 which was on a really old branch. This PR goes to 3.2 instead of master because master was breaking in CI.

JSMike:
Had to fix head for PR

PR #656, opened 2016-08-04
Implement sandbox environment to the test API

### LINKED ISSUE #6: Test sandbox
Provide an opt-in sandbox interface for tests that gives them a fresh window context to do things to.

hakatashi:
This fixes #6.

This implements the “sandbox environment” to the Intern's Test API, both in node.js and browser. @dylans

The sandbox isolates any changes to the environment, such as the global variables or native objects. And to make the test to the application available, I also implemented the fixture API to bring any variables into the sandbox. The fixtures are deeply cloned each time the test is executed and changes to the fixtures does not affect the host environment.

These API can be accessed through the Test.sandbox [boolean] and Test.fixtures [object] property.

Example:

```js<br />const test = new Test({<br />sandbox: true,<br />fixtures: {<br />assert: assert,<br />foo: 'bar',<br />},<br />test: () => {<br />assert(foo === 'bar');<br />},<br />});<br /><br />test.run();<br />``` Possible further works:
- Complete the API to the asynchronous tests: Currently only the async tests by Promise are supported and `this.async()` doesn't work. This may be fixed.
- Write documentation
- Add more types of objects that can be cloned: Though cloning object instances is impossible in JavaScript by design, it will be reasonable to add support for simulated clones of some native classes. Currently Date and RegExp and Buffer are supported. Adding more types of objects will enhance the performance.

### DIFF
--- a/lib/Test.js
+++ b/lib/Test.js
@@ -1,7 +1,9 @@
 define([
 	'dojo/Promise',
+	'dojo/has',
+	'dojo/has!host-node?dojo/node!child_process',
 	'./util',
-], function (Promise, util) {
+], function (Promise, has, child_process, util) {
 	function Test(kwArgs) {
 		for (var k in kwArgs) {
 			this[k] = kwArgs[k];
@@ -20,6 +22,7 @@ define([
 		hasPassed: false,
 		skipped: null,
 		error: null,
+		sandbox: false,
 
 		/**
 		 * The unique identifier of the test, assuming all combinations of suite + test are unique.
@@ -178,7 +181,20 @@ define([
 
 			return start()
 				.then(function () {
-					var result = self.test();
+					var result;
+
+					if (self.sandbox) {
+						if (has('host-node')) {
+							var subprocess = child_process.spawn(process.argv[0], ['-e', self.test.toString()]);
+							subprocess.stdout.pipe(process.stdout);
+							subprocess.stderr.pipe(process.stdout);
+							result = new Promise(function (resolve, reject, progress, setCanceler) {
+								subprocess.on('close', resolve);
+							});
+						}
+					} else {
+						result = self.test();
+					}
 
 					// Someone called `this.async`, so this test is async; we have to prefer one or the other, so
 					// prefer the promise returned from the test function if it exists, otherwise get the one that was
--- a/lib/Test.js
+++ b/lib/Test.js
@@ -196,25 +196,30 @@ define([
 							result = self.test();
 						} else {
 							var dfd = util.createDeferred();
+
 							var newIframe = document.createElement('iframe');
 							newIframe.width = 500;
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: lib/Test.js [hunk 1]
```diff
@@ -215,6 +215,9 @@ define([
 								// Create <script> element to store the test code
 								var script = newIframe.contentWindow.document.createElement('script');
 								script.type = 'text/javascript';
+
+								// With old IEs all errors should be catched with try-statement
```
agcolom: Hi, I just spotted a small typo: catched -> caught

hakatashi: Thanks, I'll fix this...

rozaev:
@hakatashi @morrinene @agcolom @dylans
Hello everyone! 
Right now I am write the same sandbox for my project. This pull request is actual? Why it didn't merged?

@dylans is there are any chance to see that feature in intern 3?

In my case I use sandbox only for executing application code, it's looks like that:

```JavaScript
var page, module;

bdd.beforeEach(function () {
	page = new PageWrapper();

	return page.loadModule("url/to/module").then(function (loadedModule) {
		module = loadedModule;
	});
});

bdd.afterEach(function () {
	page.destroy();
});

bdd.it("module checks", function () {
        expect(module.foo()).be.equal(42);
});
```

In fact, PageWrapper just create iframe and pass coverage variable:

``` JavaScript
var iframe = document.createElement("iframe");

iframe.contentWindow.__internCoverage = window.__internCoverage = window.__internCoverage || {};
```

rozaev:
Also PageWrapper allows to test such code (very artificial example, but still):

``` JavaScript
define([
  "dojo/xhr",
], function (
  xhr
) {
  xhr("/some/url").then(function () {
    // branch 1
  }).otherwise(function(){
    // branch 2
  });
});
```

Right now there are no way to run this code twice and check booth branches. This code will be execute only once - when we load this AMD module for first time.

``` JavaScript
var page, module;

bdd.beforeEach(function () {
	page = new PageWrapper();
});

bdd.afterEach(function () {
	page.destroy();
});

bdd.it("branch 1", function () {
        page.server.responseWith("");
        return page.loadModule("url/to/module").then(function () {
           // check branch 1 logic
        });
});

bdd.it("branch 2", function () {
        return page.loadModule("url/to/module").then(function () {
           // check branch 2 logic
        });
});
```

Of course it's a very strange use-case, but it's exist.
```

dylans:
@rozaev I would say it's mostly that we started working on Intern 4 which sort of took our focus away from this and so we've sadly not really had time to test and verify it sufficiently.

If you can test and verify that it works well with Intern 3.x, then I think we could land it for the 3.x branch and have it be part of what will at some point be a 3.5 release that rolls up any remaining 3.x features.

jason0x43:
Closing this since it's been overcome by events.

PR #704, opened 2016-12-19
Moved @types into devDependencies in package.json (from dependencies)

andfaulkner:
Was there a specific reason for having @types/node and @types/chai in package.json's dependencies rather than devDependencies? Having them in package.json's dependencies causes compiler warnings/errors when building Typescript projects (version 2 and up tested) that are not using node_modules/@types to store type definitions. Manually removing the folders gets rid of the errors, but every time a new package is installed, they return.

Type definitions should be in devDependencies anyway, especially for definitions as widely used as Node's.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,12 +22,12 @@
     "lodash-amd": "4.16.4",
     "mimetype": "0.0.8",
     "platform": "1.3.1",
-    "source-map": "0.1.33",
-    "@types/chai": "3.4.34",
-    "@types/node": "6.0.48"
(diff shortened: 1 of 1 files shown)


jason0x43:
They are in dependencies because they're dependencies, not just dev dependencies. I mean, assuming you're using Intern in a TS project, you will need those types, and @types packages should be the way to handle those in a project. I can see how that would cause problems if you're using your own types, though. Does adding the @types to an `exclude` block in your tsconfig work?

PR #713, opened 2017-01-26
Add currentTest to suite

LingSpb:
It's good to know which test is currently running, so we can interact with it, for example, in beforeEach/afterEach

### DIFF
--- a/lib/Suite.js
+++ b/lib/Suite.js
@@ -39,6 +39,8 @@ define([
 		 */
 		publishAfterSetup: false,
 
+		_currentTest: null,
+
 		/**
 		 * A flag used to indicate whether a test run shoudl stop after a failed test.
 		 */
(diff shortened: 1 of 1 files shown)


jason0x43:
`beforeEach` and `afterEach` are already passed the current Test object. I think this style of access (passing the current test Test to potentially interested parties) may be preferable to having a class-level reference since it will continue to work if, say, Suites are able to run tests concurrently.

LingSpb:
> beforeEach and afterEach are already passed the current Test object.

I don't understand well what you meant here. Actually, my original purpose was something like "in beforeEach: if the test name is in patternA, then do this, if patternB then do that..", or: "in afterEach: if the test fails, then do this..". Is any other way I can do it?

> Suites are able to run tests concurrently.

As I understand, if we use different sessions, then it should be fine (it is what I use and already tested)

Or maybe you have a better solution?

jason0x43:
Since `beforeEach` and `afterEach` are passed a reference to the test, you can already do what you were trying to do:

```js
beforeEach: function (test) {
    if (/abc/.test(test.name)) {
        ...
    } else if (/def/.test(test.name)) {
        ...
    }
},

afterEach: function (test) {
    if (test.error) {
    }
},
```

> As I understand, if we use different sessions, then it should be fine (it is what I use and already tested)

Yes, currently the only level of concurrency generally used by Intern is running sessions in parallel. I just meant that if we wanted to run the tests themselves in parallel in a session (assuming we had dealt with the other relevant issues there), a single 'currentTest' parameter wouldn't work since multiple tests in a suite could be active at once.

LingSpb:
I got it, thank you very much, @jason0x43 !

PR #719, opened 2017-02-10
Unpin dependencies.

sholladay:
Intern has previously had a very inconsistent use of pinned and unpinned dependencies. This PR aims to normalize that and improve the user experience when upstream features are released that do not require code changes to Intern (e.g. new tunnels or env vars added to [DigDug](https://github.com/theintern/digdug)).

The basic premise here is that whatever gain is had from pinning the immediate dependencies is cancelled out by virtue of the nested dependencies using ranges themselves (extremely common, in part due to `npm` defaults).

So in effect, prior to this PR, Intern still experiences the dangers of version ranges, but without the benefits. I think Intern should either [shrinkwrap](https://docs.npmjs.com/cli/shrinkwrap) its dependencies or use this PR.

I left `dojo` alone since it is pinned to a pre-release version. Using `^2.0.0-alpha.7` is valid, but since it matches both pre-releases and releases, if a breaking change happens prior to a proper release, that would be problematic, which seems within the realm of possibility (and is allowed by [the spec](http://semver.org/#spec-item-9)).

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,24 +10,24 @@
   },
   "license": "BSD-3-Clause",
   "dependencies": {
-    "benchmark": "2.1.1",
-    "chai": "3.5.0",
-    "charm": "0.2.0",
-    "diff": "1.1.0",
-    "digdug": "~1.6.0",
+    "benchmark": "^2.1.1",
+    "chai": "^3.5.0",
+    "charm": "^0.2.0",
+    "diff": "^1.1.0",
+    "digdug": "^1.6.0",
     "dojo": "2.0.0-alpha.7",
-    "glob": "7.0.3",
-    "istanbul": "0.4.1",
-    "leadfoot": "~1.7.0",
-    "lodash-amd": "4.16.4",
-    "mimetype": "0.0.8",
-    "platform": "1.3.1",
-    "source-map": "0.1.33",
-    "@types/chai": "3.4.34",
-    "@types/node": "6.0.48"
+    "glob": "^7.0.3",
+    "istanbul": "^0.4.1",
+    "leadfoot": "^1.7.0",
+    "lodash-amd": "^4.16.4",
+    "mimetype": "^0.0.8",
+    "platform": "^1.3.1",
+    "source-map": "^0.1.33",
+    "@types/chai": "^3.4.34",
+    "@types/node": "^6.0.48"
   },
   "devDependencies": {
-    "intern": "3.2.3"
+    "intern": "^3.2.3"
   },
   "scripts": {
     "install": "node support/fixdeps.js",


jason0x43:
I would say that Intern takes a reasonably consistent approach to its use of pinned dependencies; originally all dependencies were pinned, and currently all non-Intern dependencies are pinned. (Dig Dug and Leadfoot were unpinned so that bug fixes could come through without requiring a new Intern release.) I agree that only pinning top-level dependencies isn't particularly thorough, but while unpinning them may provide quicker access to new features, it magnifies the likelihood that something will break (as the CI test run shows).

At the moment I lean towards shrinkwrap, or at most using '~'.

vladikoff:
> At the moment I lean towards shrinkwrap, or at most using '~'.

👍

sholladay:
> all non-Intern dependencies are pinned

Fair. It is consistent in that sense. I would ask, though, if you trust updates from yourself, why not new features?

> it magnifies the likelihood that something will break (as the CI test run shows).

I haven't looked deep into the root cause of the Travis failure, but it failed only on Node 0.12 with a simple `AssertionError`, indicating that there may be a bug. This doesn't look like semver abuse or anything like that. And again, yes, new bugs can happen occasionally with or without this PR. I would argue that this PR makes it more obvious _why_ that can happen. It embraces a much more common convention. Another way to look at this is, which do you think is more common: new bugs OR (bug fixes AND new features)? Remember that we are talking about production releases here that have passed their respective test suites and CI runs, as all of Intern's immediate dependencies are high-quality projects that have these.

In regards to stability, note that many of your fellow testing frameworks such as [AVA](https://github.com/avajs/ava/blob/45fec330cb1fc117d848efe08fc0359814ce70a8/package.json#L89) and [Mocha](https://github.com/mochajs/mocha/blob/a2fc76c1c0e78955763ab29079792a3178eaf9bd/package.json#L321) use caret ranges and don't shrinkwrap. While I wouldn't use that as a reason alone to do this, I think it's meaningful to demonstrate that other stability-minded folks are getting along just fine with this.

jason0x43:
> This doesn't look like semver abuse or anything like that. And again, yes, new bugs can happen occasionally with or without this PR. I would argue that this PR makes it more obvious why that can happen.

I'm not so concerned with why something is broken so much as the fact that it's broken. With pinned dependencies, we (theoretically) know that users will end up with a product that has been tested. When we use '~', the odds go up that a user will encounter something broken, but those odds should still be reasonably low since '~' implies small changes. With '^' the odds of encountering breaks go up again, because much larger changes are allowed.

Noticing that Intern is broken isn't too bad -- we can just run a daily or weekly CI cron job to check that current installs work. However, dealing with the problem becomes much more of an issue, because now all new installs of Intern are broken until we do something about it (either pinning to a known good version of a dependency or writing around the issue).

> Another way to look at this is, which do you think is more common: new bugs OR (bug fixes AND new features)? Remember that we are talking about production releases here that have passed their respective test suites and CI runs, as all of Intern's immediate dependencies are high-quality projects that have these.

My main concern isn't that the projects aren't tested, it's that they aren't necessarily tested in the same environments that Intern is. The CI failure is a good example of this. At least one of the dependencies wasn't tested in, or may explicitly no longer support, Node 0.12, which Intern still technically does.

> In regards to stability, note that many of your fellow testing frameworks such as AVA and Mocha use caret ranges and don't shrinkwrap. While I wouldn't use that as a reason alone to do this, I think it's meaningful to demonstrate that other stability-minded folks are getting along just fine with this.

I'm not completely against using looser dependency versions, but I'm still not sure that the benefits necessarily outweigh the accompanying maintenance headaches.

sholladay:
Closing for lack of interest combined with merge conflicts and CI failure. 👻

sholladay:
Closing for lack of interest combined with merge conflicts and CI failure. 👻

PR #730, opened 2017-04-12
Allow equals sign in arg string

rvna:
Currently, there is no way to pass a string containing an equals sign into intern.args. This might be necessary sometimes, for example when running tests with a protected URL token. 

Behavior before change: Everything before the first equals sign in an arg is the name, the substring between the first and second equals sign is the value
Behavior after change: Everything before the first equals sign in an arg is the name, everything after the first equals sign is the value


alexre0:
Would it still be possible to merge this in?
@jsf-clabot

jason0x43:
Definitely. I keep forgetting that JS's `split` always splits on every instance of a match, even when you specify a limit. :)

alexre0:
Thanks!

PR #753, opened 2017-06-16
Update HowTo.md and running.md

CodeTroopers:
config function doesn't exist. the function is configure


jason0x43:
Good catch! I made a couple minor changes and merged this in 20ccd155.

PR #778, opened 2017-07-22
Include coverage sources with glob patterns in Intern 4

bryanforbes:
This removes `coverageSources`, deprecates `excludeInstrumentation`, and adds `coverage` to the Intern config. It also adds the functionality to negate glob patterns in an array of patterns.

### DIFF
--- a/intern.json
+++ b/intern.json
@@ -96,8 +96,6 @@
 
 		"coverage": {
 			"description": "Generate full coverage",
-			// "benchmark": true,
-			// "extends": [ "allSuites" ]
 			"coverage": [
 				"_build/src/lib/**/*.js",
 				"_build/src/bin/**/*.js",
--- a/intern.json
+++ b/intern.json
@@ -95,13 +95,15 @@
 		},
 
 		"coverage": {
+			"benchmark": true,
 			"description": "Generate full coverage",
 			"coverage": [
 				"_build/src/lib/**/*.js",
 				"_build/src/bin/**/*.js",
 				"_build/src/loaders/**/*.js",
 				"_build/src/tasks/**/*.js"
-			]
+			],
+			"extends": [ "allSuites" ]
 		},
 
 		"wd": {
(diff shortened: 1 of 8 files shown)


intern.json
```diff
@@ -88,24 +87,21 @@
...
+		"progress": {
+			"description": "For showing test progress",
 			"node": {
-				"reporters": { "name": "runner", "options": { "hideSkipped": true, "hidePassed": true } }
+				"reporters": { "name": "runner", "options": { "hideSkipped": true, "hidePassed": false } }
 			}
 		},
 
 		"coverage": {
 			"description": "Generate full coverage",
```
jason0x43: We should put back the `extends` and `benchmark` properties, or add them to CI. I just commented them out to make dev faster.

src/lib/executors/Node.ts
```diff
@@ -477,6 +490,22 @@ export default class Node extends Executor<Events, Config, NodePlugins> {
 				this._setOption(name, parseValue(name, value, 'object[]', 'browserName'), addToExisting);
 				break;
 
+			case 'excludeInstrumentation':
+				this.emit('deprecated', {
+					original: 'excludeInstrumentation',
+					replacement: 'instrument'
```
jason0x43: Should be `coverage`

tests/support/unit/executor.ts
```diff
@@ -13,7 +13,9 @@ export function testProperty<E extends Executor = Executor, C extends Config = C
 ) {
 	assert.throws(() => { executor.configure(<any>{ [name]: badValue }); }, error);
 	executor.configure(<any>{ [name]: goodValue });
-	assert.equal(mockConsole.warn.callCount, 0, 'no warning should have been emitted');
+	for (let call of mockConsole.warn.getCalls()) {
```
jason0x43: We should be passing something in to testProperty indicating that a deprecation warning (or any other console output) is expected.

jason0x43:
Merged in 694a247d4eb08acde64c6b57f76c159a8d0006fb

PR #781, opened 2017-08-03
Fix coverage bugs

bryanforbes:
Currently, full coverage (including uncovered files) only displays for the final report if there are more than two sessions. This PR contains changes to output full coverage for all sessions.

### DIFF
--- a/src/lib/executors/Node.ts
+++ b/src/lib/executors/Node.ts
@@ -2,7 +2,7 @@ import Executor, { Config as BaseConfig, Events as BaseEvents, Plugins } from '.
 import Task, { State } from '@dojo/core/async/Task';
 import { parseValue, pullFromArray } from '../common/util';
 import { expandFiles, normalizePath, readSourceMap } from '../node/util';
-import { stat, readFile, readFileSync } from 'fs';
+import { readFileSync } from 'fs';
 import { deepMixin, mixin } from '@dojo/core/lang';
 import ErrorFormatter from '../node/ErrorFormatter';
 import { dirname, normalize, relative, resolve, sep } from 'path';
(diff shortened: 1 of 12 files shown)


jason0x43:
Some thoughts:

* Since reference sharing is an issue, we may want to:
  1. duplicate the event data for every handler we call in `emit` (easy, but not the most efficient thing)
  2. require coverage data to be duplicated before merging rather than modifying the data in the executor's 'coverage' listener (more efficient, but you have to remember)
* We should consider moving initial coverage generation to some point after testing (maybe after unit tests) for efficiency. Right now we spend a significant amount of time before testing even starts generating initial coverage data, then we're going to do it again for any files loaded by suites. We should only generate initial data for files that aren't loaded by suites (i.e., that don't have any coverage after testing is finished).

jason0x43:
Also, rather than having Node's listener merge incoming data with initial coverage, we may want to just have the code that generates initial coverage emit 'coverage' events with that data.

src/lib/executors/Node.ts
```diff
@@ -127,6 +147,7 @@ export default class Node extends Executor<Events, Config, NodePlugins> {
 
 		this.on('coverage', message => {
 			this._coverageMap.merge(message.coverage);
+			message.coverage = mergeCoverageData(duplicate(this._initialCoverage), message.coverage);
```
jason0x43: Rather than altering message data for other listeners, maybe we should:

1. do the duplication in `emit`
2. emit the initial coverage with its own 'coverage' event rather than merging new data into it here

src/lib/executors/Node.ts
```diff
@@ -269,6 +290,14 @@ export default class Node extends Executor<Events, Config, NodePlugins> {
 	}
 
 	protected _beforeRun() {
+		this._coverageFiles.forEach(filename => {
```
jason0x43: This adds quite a startup delay, and ends up instrumenting any files being tested twice. Could we do it at the end of unit tests, and only for files that didn't already have coverage data?

morrinene:
#784

Issue #724, opened 2017-03-21
Improve HTML reporter


jason0x43:
Suggested improvements:

- Ability to hide passing tests (so that only failing tests are shown), and/or some way to jump between failing tests (useful for very large suites)
- Ability to click on a failing test and have Intern run just that test (can be implemented with a link containing a specific `grep` query argument)

jason0x43:
This was addressed in https://github.com/theintern/intern/commit/bec9d5792dd4a8843d501a7b14d1daae74c6d9a8

PR #801, opened 2017-09-25
Improve HTML reporter #724

rozaev:
Hello @dylans, hello @jason0x43!

I decide to help with https://github.com/theintern/intern/issues/724 issue.

There is an initial pull request.

Known problems:

1. Icon for skipped tests is not exist. 
There is a `$skippedIcon` variable with `\e807` value in `intern/lib/reporters/html/variables.styl`, but fonts don't contain that character.

2. I don't know how link should look like.  Any ideas / markup / CSS class? 

3. Need to write functional tests for that path.

Now it's looks like this:
screenshot from 2017-09-25 20-13-23

### DIFF
--- a/lib/reporters/Html.js
+++ b/lib/reporters/Html.js
@@ -1,7 +1,40 @@
 define([
 	'require',
-	'../util',
-], function (require, util) {
+
+	'../util'
+], function (
+	require,
+
+	util
+) {
+	var location = window.location;
+
+	function getFullName(test) {
+		var name = test.name;
+		while (test.parent && test.parent.name) {
+			test = test.parent;
+			name = test.name + ' - ' + name;
+		}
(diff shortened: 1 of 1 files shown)


jason0x43:
Thanks for your interest! Could you rework this against the master branch? (The changes should be pretty similar there.) We'd prefer to put new features in the Intern 4 code, at least for now.

jason0x43:
As far as the link styling goes, we can leave that at the default for now.

rozaev:
Okay, now this pull request against master branch.

There are also another problem: I use ` window.location` in my patch, but it is not exist in nodejs environment. How I can check in which environment I am?

rozaev:
Rework against master branch: https://github.com/theintern/intern/pull/802

Issue #636, opened 2016-05-20
Feature Request: Retry test in browser client


odoe:
It would be great if each test in the browser client had a button/link to rerun only that test. Similar to the QUnit client.

This is incredibly useful when working on 1-2 test items at a time.

sholladay:
Out of curiosity, how long do your unit tests take to run? Is it really more convenient to go hunt down a link with your mouse and click it vs Command + Shift + R on your keyboard?

odoe:
All our suites take about 10 minutes. It's not a huge deal while working on a particular feature. I can typically isolate the tests to a single suite. But I still have to test everything to make sure I don't introduce bugs or break anything. That's when it comes in handy to try and nail down those failing tests.

RoystonS:
The other use case for this feature is unrelated to execution speed: if you want to put a breakpoint in your code to see what happens during a failing unit test: re-running all the tests makes that problematic as the breakpoint will be hit for multiple running tests, not just the one you're interested in.  It's very useful to be able to focus in on one or a few unit tests.  Right now we end up going back to the code, commenting out (or skipping) uninteresting tests and running it again to ensure that if the breakpoint is hit, it's only during the interesting test.

sholladay:
> our suites take about 10 minutes

But presumably that is mainly due to functional suites. Yes? The browser client doesn't deal with those. So this would only affect unit tests, which are usually very fast.

> It's very useful to be able to focus in on one or a few unit tests

Amen. I think https://github.com/theintern/intern/issues/514 is a better solution, though. E.g. it is generic (isn't specific to unit tests) and is probably still faster to work with, since you don't have to scroll a huge test result table and hunt for a button.

jason0x43:
This should be pretty easy to do by just giving the report items targeted links. Like, each suite item would have a link like `<existing URL>&suites=<mid>`. Test items would have links pointing to `<existing URL>&suites=<mid>&grep="test id"`. Clicking the link would run only that suite or test. The user could use the browser's back button to run all tests again.

jason0x43:
This was handled by #802, which makes test ID's clickable links in the HTML reporter.

PR #802, opened 2017-09-26
improve HTML - init commit

rozaev:
Hello @dylans, hello @jason0x43!

This pull request can fix several issues:
https://github.com/theintern/intern/issues/636  https://github.com/theintern/intern/issues/724

Now it's looks like this:
screenshot from 2017-09-25 20-13-23

Feedback:
`npm test serveOnly` command say `Listening on localhost:9000 (ws 9001)` it's not comfortable - maybe it's should say `Listening on  http://127.0.0.1:9000/__intern/  (ws 9001)`? In that case you can copy-paste this URL to browser.

If you prefer arrow function for callback - maybe you should add linter rule for that?
https://eslint.org/docs/rules/prefer-arrow-callback

### DIFF
--- a/src/lib/reporters/Html.ts
+++ b/src/lib/reporters/Html.ts
@@ -11,6 +11,8 @@ export default class Html extends Reporter implements HtmlProperties {
 
 	document: Document;
 
+	location: Location;
+
 	protected _reportContainer: Element;
 
 	// Div element to hold buttons above the summary table
@@ -57,6 +59,7 @@ export default class Html extends Reporter implements HtmlProperties {
 	constructor(executor: Browser, options: HtmlOptions = {}) {
 		super(executor, options);
 		this.document = options.document || window.document;
+		this.location = options.location || window.location;
 		this._fragment = this.document.createDocumentFragment();
 	}
 
@@ -323,7 +326,7 @@ export default class Html extends Reporter implements HtmlProperties {
 		this._suiteCount++;
 
 		cellNode.className = 'title';
-		cellNode.appendChild(document.createTextNode(suite.name));
+		cellNode.appendChild(this.createLinkNode(suite));
 		rowNode.className = 'suite';
 		rowNode.appendChild(cellNode);
 		this._reportNode.appendChild(rowNode);
@@ -466,7 +469,7 @@ export default class Html extends Reporter implements HtmlProperties {
 			cellNode.className += ' indent' + this._indentLevel;
 		}
 
-		cellNode.appendChild(document.createTextNode(test.name));
+		cellNode.appendChild(this.createLinkNode(test));
 		rowNode.appendChild(cellNode);
 
 		cellNode = document.createElement('td');
@@ -499,10 +502,28 @@ export default class Html extends Reporter implements HtmlProperties {
 
 		this._reportNode.appendChild(rowNode);
 	}
+
+	private createLinkNode(obj: Suite | Test) {
+		const document = this.document;
+		const location = this.location;
+		const a = document.createElement('a');
+		let search = location.search;
+
+		if (search) {
+			search = search.slice(1).split('&').filter(el => {
+				return !el.startsWith('grep');
+			}).join('&');
+		}
+
+		a.href = location.origin + location.pathname + `?${search}&grep=${encodeURIComponent(obj.id)}`;
+		a.appendChild(document.createTextNode(obj.name));
+		return a;
+	}
 }
 
 export interface HtmlProperties extends ReporterProperties {
 	document: Document;
+	location: Location;
 }
 
 export type HtmlOptions = Partial<HtmlProperties>;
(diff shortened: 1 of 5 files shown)


src/lib/reporters/Html.ts
```diff
@@ -3,6 +3,32 @@ import Reporter, { eventHandler, ReporterProperties } from './Reporter';
 import Test from '../Test';
 import Suite from '../Suite';
 
+const location = window.location;
+
+function getFullName(test: Suite | Test): string {
```
jason0x43: Don't specify a return type when it's not necessary.

src/lib/reporters/Html.ts
```diff
@@ -3,6 +3,32 @@ import Reporter, { eventHandler, ReporterProperties } from './Reporter';
...
+	}
+	return name;
+}
+
+function createLinkNode(test: Suite | Test): Element {
+	const a = document.createElement('a');
+	let search = location.search;
+
+	if (search) {
+		search = search.slice(1).split('&').filter(function (el) {
```
jason0x43: For consistency, prefer arrow functions for simple callbacks.

jason0x43:
I like where this is going. Some thoughts:

* Use `suites` instead of `grep` (or both `suites` and `grep` if a specific test is being rerun) so that the only result in the re-run report will be the suite the user selected.
* Allow the location to be passed into the reporter, like the `document` property. Add it to `HtmlProperties` and make it a reporter property, then move `createLinkNode` and `getFullName` into Html as methods, or allow the location to be passed into `createLinkNode`. This will improve testability.

jason0x43:
Also, the 'skipped' icon is rendering fine for me using Chrome on a Mac. Is there a particular browser you're seeing the issue with?

src/lib/reporters/Html.ts
rozaev: done

src/lib/reporters/Html.ts
rozaev: done

rozaev:
> Also, the 'skipped' icon is rendering fine for me using Chrome on a Mac. Is there a particular browser you're seeing the issue with?

It was my mistake - I use outdated master branch.

rozaev:
> Use suites instead of grep (or both suites and grep if a specific test is being rerun) so that the only result in the re-run report will be the suite the user selected.

A you sure about that? We can't map program suites with suites param.
Suites param is array of string like `path/to/suite1.js, ...etc`. Program suites has specific name, like `bdd.describe("My first suite!", () => {});`. Seems that I can't just correlate them.

Maybe we should build another html representation in case when grep property is specified?

rozaev:
@jason0x43 one more question. I tried to write tests but looks like i should register me mocks somehow. There is an error:

```
(ノಠ益ಠ)ノ彡┻━┻
Error: A plugin named "createLocation" has not been registered
  at Node.Executor.getPlugin  <src/lib/executors/Executor.ts:322:10>
  at <tests/unit/lib/reporters/Html.ts:6:31>
  at Object.defineProperty.value  <_tests/tests/unit/lib/reporters/Html.js:3:17>
  at Object.<anonymous>  <_tests/tests/unit/lib/reporters/Html.js:9:3>
```

jason0x43:
Assuming you want to use the same createLocation code in the browser and in Node, you don't need to deal with the plugin system, just do `import { createLocation } from '../../../support/unit/mocks';`. The setup used for DOM documents involves jsdom, and is a bit more complex than that to keep Intern from trying to load jsdom in browsers (or use the global `document` in Node).

jason0x43:
So, how's this going? From the last CI error, it looks like the self tests are just missing a few commas.

rozaev:
Well, I believe that feature itself is ready. There are only one little problem: I add test for `testEnd` method - now it's check that link for test exist. 

But I can't just add test for `suiteStart` method because it's didn't actually add link anywhere. It's just store it in private property: `this._runningSuites[suite.id] = { node: rowNode };`

So - only one unit test left. Seems that I should call multiply methods if I want test this. It's sad - I was sure that I can call only `suiteStart`.

jason0x43:
You could aspect `document.createElement` for the reporter's document, and verify whether the expected elements were created or not, maybe something like:

```js
import { after } from '@dojo/core/aspect';

// ...

test() {
    let nodes: HTMLElement[] = [];
    after(reporter.document, 'createElement', (returnValue: any, args: IArguments) => {
        nodes.push(returnValue);
        return returnValue;
    });

    // do test

    // make assertions about created nodes
}
```

rozaev:
@jason0x43 It's ready now.

rozaev:
@jason0x43 any news?

jason0x43:
Hah, good timing, I was just reviewing it.

jason0x43:
I made a couple minor changes (updated the createLinkNode method to use `@dojo/core/UrlSearchParams` and styled the URLs to use the text colors). Thanks for your contribution!

PR #813, opened 2017-10-10
Update Grunt task

### LINKED ISSUE #812: Support config files in the grunt task
The grunt task currently assumes all config will come from a grunt options object. Support a `config` grunt option that works like the `config` option for the Node runner (i.e., it will accept something of the format '[file.json][@config]').

bryanforbes:
* Allow grunt options to specify 'config' to point to a config file and/or section
* Change node/util.getConfig to accept an optional second argument to allow passing in other command line arguments
* Updated task tests

Fixes #812

### DIFF
--- a/src/tasks/intern.ts
+++ b/src/tasks/intern.ts
@@ -1,26 +1,46 @@
 import global from '@dojo/shim/global';
+import { assign } from '@dojo/core/lang';
 
 import Node from '../lib/executors/Node';
 import { Config } from '../lib/executors/Executor';
+import { getConfig } from '../lib/node/util';
 
 interface TaskOptions extends grunt.task.ITaskOptions, Partial<Config> {
 	[key: string]: any;
 }
 
+function configure(options: TaskOptions): Promise<{
+	config: Partial<Config>;
+	options: TaskOptions;
+}> {
+	if (options.config) {
+		return getConfig(options.config, []).then(({ config }) => {
+			delete options.config;
+
+			return { config, options };
+		});
+	}
+
+	return Promise.resolve({ config: {}, options });
+}
+
 export = function(grunt: IGrunt) {
 	grunt.registerMultiTask('intern', function() {
 		const done = this.async();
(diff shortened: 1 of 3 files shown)


src/tasks/intern.ts
```diff
@@ -1,26 +1,46 @@
...
 
 import Node from '../lib/executors/Node';
 import { Config } from '../lib/executors/Executor';
+import { getConfig } from '../lib/node/util';
 
 interface TaskOptions extends grunt.task.ITaskOptions, Partial<Config> {
 	[key: string]: any;
 }
 
+function configure(options: TaskOptions): Promise<{
```
jason0x43: `configure` is a weird name for this since its not configuring anything (although I don't suppose it matters since its internal). Also, Intern code typically has things like module-level constants and functions declare below exports.

src/tasks/intern.ts
```diff
@@ -1,26 +1,46 @@
...
+	config: Partial<Config>;
+	options: TaskOptions;
+}> {
+	if (options.config) {
+		return getConfig(options.config, []).then(({ config }) => {
+			delete options.config;
```
jason0x43: Instead of deleting config from the passed-in options, it'd be nicer to return a new object that didn't have the config property (since we're returning options anyway).

tests/unit/tasks/intern.ts
```diff
@@ -46,15 +55,15 @@ registerSuite('tasks/intern', function() {
 		},
 
 		beforeEach() {
-			mockRun.reset();
+			sandbox.reset();
 			mockRun.resolves();
```
jason0x43: If we made `mockRun = sandbox.stub().resolves()`, could we get buy with just calling `sandbox.resetHistory()` above (instead of `sandbox.reset()`)?

bryanforbes: There are a few other stubs I've set up that need to be reset every time

jason0x43:
Merged in def6346

PR #842, opened 2017-11-16
Fix plugin options

xdissent:
This patch sets the loading plugin options at the appropriate time. Right now they're getting wiped out before the loader has a chance to run.


jason0x43:
Good catch, thanks!

PR #879, opened 2018-02-08
Added more documentation for configuring plugins and loaders

devpaul:


docs/configuration.md
```diff
@@ -313,6 +315,65 @@ Given the config above, and assuming child config “d” is selected, the follo
 3. The result of 2 will be mixed into the base config
 4. The result of 3 will be the resolved config
 
+## Configuring loaders
+
+A loader is a script used by Intern to load suites into the testing environment. Only a single loader may be specified per
+ environment (Node or browser). Intern has a number of [loaders](https://github.com/theintern/intern/tree/master/src/loaders)
```
jason0x43: This implies (to me) that Intern actually comes with various loaders. It might be better to say that Intern can _work with_ a number of loaders.

docs/configuration.md
```diff
@@ -313,6 +315,65 @@ Given the config above, and assuming child config “d” is selected, the follo
...
+ available. If no loader is defined, then the default loaders are used per environment.
```
jason0x43: Maybe make it clearer that "default loader" will mean "no loader" in the browser, which means no modules.

docs/configuration.md
```diff
@@ -313,6 +315,65 @@ Given the config above, and assuming child config “d” is selected, the follo
...
+            "script": "tests/support/dojoMocking.js",
+            "useLoader": true
+        }
+    ]
+}
+```
+
+Plugin configurations with `options` are passed to the plugin when it is registered with Intern. By default, plugins are 
+ loaded before an external loader is available. Plugins with the `useLoader` option set will be loaded later after a 
+ loader becomes available.
```
jason0x43: Maybe clarify the implications here a bit more. Plugins loaded before an external loader are available _must_ be standalone scripts in the browser because there's no loader. Plugins with `useLoader` aren't just loaded after the loader becomes available but _with_ the loader, meaning they can be modules.

docs/configuration.md
```diff
@@ -313,6 +315,65 @@ Given the config above, and assuming child config “d” is selected, the follo
...
+```
+
+It's useful to think of Intern's loaders as glue code that Intern uses to load configured suites within the environment.
+ Consequently, most of the provided Intern loaders also require the loader package to be installed in `node_modules` 
+ using NPM. Look at the [loader code](https://github.com/theintern/intern/tree/master/src/loaders) for more options and
+ requirements.
+
+## Configuring plugins
+
+Plugins are standalone scripts that are loaded by Intern. Any number of them may be loaded in-order following a similar
```
jason0x43: They can also be modules if `useLoader` is true (so maybe "standalone scripts or modules").

docs/configuration.md
devpaul: I can see where that's a bit confusing for a loaderless architecture. What's the terminology we're using for the "loader" scripts?

jason0x43: "Loader scripts" is what I've been using, but I'm open to suggestion.

devpaul: Loader scripts is good. I'm at a bit of a disadvantage in how to explain it because I'm familiar w/ the architecture. I think we'll be OK as long as we're consistent and take the time to highlight what a loader script does.

Issue #894, opened 2018-03-28
Support for running multiple Intern instances in a single Grunt run


rhpijnacker:
In my grunt configuration, I have split the tasks so I can run part of my tests with chrome, safari and ie11 and part with only chrome and safari.
When I run both tasks in one grunt run, the first completes succesfully. The second one, gives an error at the start:
```
Running "intern:batch2" (intern) task
Error: Loader script node_modules/intern/loaders/dojo.js did not register a loader callback
  at <src\lib\executors\Executor.ts:915:12>
  at <node_modules\@dojo\core\async\Task.ts:351:12>
  at handler  <node_modules\@dojo\core\async\ExtensiblePromise.ts:229:14>
Warning: Loader script Platform/WebApps/ISCV/interntests/dojoloader.js did not register a loader callback Use --force to continue.
```
The reason seems to be that (since the loader is already required once) the Executor does not get the right trigger that the loader finished loading.

Intern version: 4.1.5
Node version: 6.11
NPM version: 5.2
Grunt version: 1.0.1

jason0x43:
Hmmm... That may require an update to the loader script API so that Intern can teardown the loaders when it finishes a run. That, or loader scripts can check for the existence of their loader during initialization and reuse or replace it (probably replace).

rhpijnacker:
At first I thought the latter would be trivial to implement, until I realized that the loader script is actually just called only once. So "loader scripts can check for the existence of their loader during initialization" would not actually work. This should be handled in `Executor`, since it is the place that loads the scripts so can administrate whether a loader was already loaded.

jason0x43:
> ...the loader script is actually just loaded and executed only once.

That, I think, is the real issue. The first time Intern is run, the loader script is loaded and it registers a loader callback. The second time Intern is run _in the same Node session_, the loader script is required but it isn’t actually re-loaded or evaluated since it’s already cached by the Node loader, so its registration function isn’t called the second time.

There are a couple ways to try to deal with this. One is for Intern to keep track of individual modules it loaded and then de-register them at the end of the session. This process could potentially be simplified a bit by adding a ‘deregistration’ function to the loader and plugin APIs that Intern could use to unload them after a run.

Another (and probably much easier in the long run) would be for Intern to load everything it needs to in a sandbox that could simply be discarded after a test run.

rhpijnacker:
Do you have a suggestion for what that deregistration function should look like?

jason0x43:
[...]

PR #898, opened 2018-03-29
configure `options` before `config` in the grunt task

rhpijnacker:
Given the following grunt config:
>     {
>         intern: {
>             subset: {
>                 ...,
>                 config: '@subset'
>             },
>             configs: {
>                   subset: { ... }
>             }
>         }
>     }

When I select the `intern:subset` task, I see the `subset` configuration being configured first (as `config`) and the more general configuration next (as `options`). By doing this in this order, the more specific `subset` configuration is overwritten, which is not how it should be.

The rationale for setting it up like this: 
- We put the actual config in `intern.json` so we can still serve it with `intern serveOnly`
- We need the `configs` to select different configurations (e.g. subsets of tests that can and cannot be loaded with IE11)
- We initialize the intern configurations from a script that loads `intern.json` and adds e.g. the `config: '@subset'` property to select the right configuration for that task.


jason0x43:
In general, the options specified in the Grunt config should take priority over the options specified in the base Intern config. Just out of curiosity, though, why does the sample Grunt config above declare it’s own ‘subset’ config?

rhpijnacker:
| Just out of curiosity, though, why does the sample Grunt config above declare it’s own ‘subset’ config?

I'm trying to set things up such that we can easily run the tests from within a browser (during development) and also from our build scripts on a local selenium grid. Since not all tests run in IE11 (due to https://github.com/rhpijnacker/intern-cucumber/issues/10) we've split them up in batches (as per your suggestion). Doing so exposes some new issues, like this one and #894.

rhpijnacker:
| In general, the options specified in the Grunt config should take priority over the options specified in the base Intern config. 

That's what I would expect too. However I'm not sure this is what I see happening now, since the config as specified via `intern:subtask` gets overwritten. How should I set up things so I can run the `subtask` as specified.

jason0x43:
If your intern config defines child configs and grunt just needs to run one of those, it should only need to specify a 'config' option with the subtask name.

```
{
    intern: {
        subset: {
            config: '@subset'
        }
    }
}
```

Other values provided in the grunt config, such as it's own version of a child config, will overwrite values loaded from the base Intern config.

I haven't had a chance to look into this yet, but if grunt isn't able to properly load child a config from an Intern config, that's a problem.

> In general, the options specified in the Grunt config should take priority over the options specified in the base Intern config.

My comment here was just to point out that this PR effectively flips the priority of the config sources, making the Intern config file have priority over the options from Grunt.

PR #955, opened 2018-10-24
Report suite errors in JUnit reporter

jonnycornwell:
Currently the JUnit reporter does not report any suite level errors in the output, resulting in the report suggesting tests have passed when suite level errors occur. The JUnit schemas I have found also suggest there is no concept of a suite level errors, see:
https://github.com/junit-team/junit5/blob/master/platform-tests/src/test/resources/jenkins-junit.xsd
https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd

This PR will update the test report to fail any tests which have not been skipped with the suite error. It is worth noting that should errors occur in the after/afterEach of tests these tests will also be marked as failed even if they have passed themselves. I feel this is a lesser of two evils as never finding out about suite errors is more misleading, there is most likely an issue which needs to be resolved.

I also set up jenkins builds with the JUnit reporter plugin to test the script below, along with the before + after report output:

```const bdd = intern.getPlugin("interface.bdd");
const { assert } = intern.getPlugin('chai');

bdd.describe("Top level", () => {
  const testCases: { context: string; setup: (cb:() => void) => void }[] = [
    { context: "a failing before", setup: bdd.before },
    { context: "a failing after", setup: bdd.after },
    { context: "a failing beforeEach", setup: bdd.beforeEach },
    { context: "a failing afterEach", setup: bdd.afterEach },
    { context: "a skipped suite", setup: _cb => bdd.before(suite => {suite.skip(`Skipped suite`)})  },
    { context: "no issues", setup: () => {} }
  ];

  testCases.forEach((testCase, i) => {
    const context = testCase.context;
    bdd.describe(`${i} Suite with ${context}`, () => {
      testCase.setup(() => {
        throw new Error(`Problem in ${context}`);
      });

      bdd.it(`${i}.1 test within a suite with ${context}`, () => {});
      bdd.it(
        `${i}.2 another test within a suite with ${context}`,
        () => {}
      );

      bdd.it(
        `${i}.3 failing test within a suite with ${context}`,
        () => {
          assert.equal(true, false, "True should be false")
        }
      );

      bdd.it(
        `${i}.4 a skipped test within a suite with ${context}`,
        test => {
          test.skip(`Skipped test: '${test.name}'`)
        }
      );
    });
  });
});
```

**Jenkins JUnit report output - BEFORE**

Only reports individual test failures

image

**Jenkins JUnit report output - AFTER**

image

### DIFF
--- a/src/lib/Suite.ts
+++ b/src/lib/Suite.ts
@@ -11,6 +11,10 @@ import Test, { isTest, SKIP } from './Test';
 import { InternError } from './types';
 import { Remote } from './executors/Node';
 
+export interface SuiteError extends InternError {
+  lifecycleMethod?: LifecycleMethod;
+}
+
 /**
  * The Suite class manages a group of tests.
  */
@@ -43,7 +47,7 @@ export default class Suite implements SuiteProperties {
   beforeEach: TestLifecycleFunction | undefined;
 
   /** The error that caused this suite to fail */
-  error: InternError | undefined;
+  error: SuiteError | undefined;
 
   /** This suite's name */
   name: string | undefined;
@@ -343,7 +347,7 @@ export default class Suite implements SuiteProperties {
     // Run the before and after suite lifecycle methods
     const runLifecycleMethod = (
       suite: Suite,
-      name: keyof Suite,
+      name: LifecycleMethod,
       test?: Test
     ): CancellablePromise<void> => {
       let result: PromiseLike<void> | undefined;
@@ -436,6 +440,7 @@ export default class Suite implements SuiteProperties {
           if (error !== SKIP) {
             if (!this.error) {
               this.executor.log('Suite errored with non-skip error', error);
+              (error as SuiteError).lifecycleMethod = name;
               this.error = error;
             }
             throw error;
@@ -476,7 +481,7 @@ export default class Suite implements SuiteProperties {
         // Run the beforeEach or afterEach methods for a given test in
         // the proper order based on the current nested Suite structure
         const runTestLifecycle = (
-          name: keyof Suite,
+          name: LifecycleMethod,
           test: Test
         ): CancellablePromise<void> => {
           let methodQueue: Suite[] = [];
@@ -560,8 +565,8 @@ export default class Suite implements SuiteProperties {
                 // An error may be associated with a deeper test
                 // already, in which case we do not want to
                 // reassociate it with a more generic parent
-                if (error && error.relatedTest == null) {
-                  error.relatedTest = <Test>test;
+                if (error && error.relatedTest == null && isTest(test)) {
+                  error.relatedTest = test;
                 }
               };
 
@@ -715,20 +720,22 @@ export default class Suite implements SuiteProperties {
       }
     });
 
-    if (this.error) {
+    const error = this.error;
+    if (error) {
       json.error = {
-        name: this.error.name,
-        message: this.error.message,
-        stack: this.error.stack
+        name: error.name,
+        message: error.message,
+        stack: error.stack
       };
 
-      if (this.error.relatedTest && this.error.relatedTest !== <any>this) {
-        // relatedTest can be the Suite itself in the case of nested
-        // suites (a nested Suite's error is caught by a parent Suite,
-        // which assigns the nested Suite as the relatedTest, resulting
-        // in nestedSuite.relatedTest === nestedSuite); in that case,
-        // don't serialize it
-        json.error.relatedTest = this.error.relatedTest.toJSON();
+      const lifecycleMethod = error.lifecycleMethod;
+      if (lifecycleMethod) {
+        json.error.lifecycleMethod = lifecycleMethod;
+      }
+
+      const relatedTest = error.relatedTest;
+      if (relatedTest) {
+        json.error.relatedTest = relatedTest.toJSON();
       }
     }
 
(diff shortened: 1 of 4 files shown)


jason0x43:
From some experimentation, both ant and maven (which generate the XML reports for junit tests) handle suite errors in a similar way -- they create a new testcase within the suite for the suite itself. So, if there's an error in a suite.before, you'd end up with

```xml
<testsuite name="suiteName">
  <testcase name="suiteName">
    <error>...</error>
  </testcase>
</testsuite>
```

Note that none of the testcases are present. When there's an error in a before (technically a `@BeforeClass` in junit), none of suite's tests are considered at all (they're not even marked ask skipped).

If there's an error in a suite.after, you'd get something like:

```xml
<testsuite name="suiteName">
  <testcase name="test1"/>
  <testcase name="test2"/>
  <testcase name="suiteName">
    <error>...</error>
  </testcase>
</testsuite>
```

All the tests pass, but there was still an error in the suite.

> This PR will update the test report to fail any tests which have not been skipped with the suite error.

There's a distinction between showing tests as failing and showing that an error occurred, and suite errors should not fail tests. I mean it's good to indicate than an error occurred; including the `error` element in a testcase does that. However, explicitly marking the test as "failed" via its `status` attribute seems less reasonable. (To be honest, I'm not sure where the `status` attribute came from; it's not in the JUnit XML reports produced by maven or ant.)

Overall,  we should probably go the same route as maven and ant and show a suite error when there's a suite error, but not show results for tests that are skipped due to suite errors, and not mark tests as failed when they haven't explicitly failed.

jonnycornwell:
Thanks for taking a look, sounds reasonable, agreed there was a bad small about failing all the tests if there was an error in the after. I wasn't too sure where status came from either but didn't want to remove it in case someone added it for a reason.

I will have a look at doing what you suggested with regards to the before and after.

jonnycornwell:
Sorry for the delay, so I have made some changes following suggestions by @jason0x43.

## In summary:

- I have added a `SuiteError` which extends `InternError` and adds the associated lifecycle method.
  - I have also centralised the typings for the lifecycle methods, see `LifecycleMethod `, using `keyof Suite` included more than just the lifecycle methods 
- If there are `before`/`beforeEach` errors then no children will be reported in that suite and will be replaced by a new `testcase` with the suite error
- If there are `after`/`afterEach` errors then a testcase with the suite error will be added after the children
- Check that the `test` (it's really a test or suite) [is actually a test](https://github.com/jonnycornwell/intern/blob/3221803b12cce200bbc8ac1204812b0e7b895970/src/lib/Suite.ts#L568) before setting the related test. This meant that a [comment + additional check](https://github.com/theintern/intern/blob/b99fe83a4816578aeb2c4cdf9f02d997e442b7c5/src/lib/Suite.ts#L725-L730) in the `toJSON` method could be removed as the relatedTest will always be a test.
- Added associated tests to suite and new tests for the updated JUnit report behaviour

## Example in jenkins:

Note: There are duplicates as the tests have been run in both node and chrome

image

## Additional comments:
- I feel it would be nicer to report `beforeEach`/`afterEach` errors on a test by test basis, rather than a suite error with an associated single test. For example, if an `afterEach` fails add an extra `testcase` after the test to represent that error or have the test itself throw with an `AfterEach` error. Changing this would also mean that there would be no need for a relatedTest on the `InternError` interface. Given these would be architecture changes I didn't really want to go down that route this this PR

jason0x43:
I looked at this a bit last week and am still reviewing it. I like separation of suite lifecycle methods from the rest of the interface, but I also agree the the current handling of `beforeEach` and `afterEach` is suboptimal. Ideally we should report the error (but not a test failure) on a test-by-test basis. I'm still poking around at that to see if there's a decent way to do that without significant rearchitecting.

jason0x43:
What if we did something like this? (Well, a blending of this and your PR.) I think it might allow us to associate lifecycle errors with tests without breaking existing code: https://github.com/theintern/intern/compare/report-suite-errors?expand=1

I'm adding test-related lifecycle errors to a `Test.suiteError` property, and recording the name of the lifecycle method on the suite error. Reporters can inspect the suite error to determine if it was test or suite-related, and they can inspect tests when there's a suite error to determine whether the test was associated with a suite error.

```
[23:33:24] ~/D/W/s/i/intern git:report-suite-errors ❮ npm --silent test suites=_tests/tests/unit/lib/Environment.js coverage= environments=node reporters=junit | xmllint --format -
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
  <testsuite name="node" failures="0" skipped="0" tests="2" time="0.002">
    <testsuite name="lib/Environment" failures="0" skipped="0" tests="2" time="0.002">
      <testcase name="constructor with info" time="0" status="0">
        <error message="test error" type="Error">Error: test error
  at Suite.afterEach  &lt;tests/unit/lib/Environment.ts:8:11&gt;
  at &lt;src/lib/Suite.ts:383:27&gt;
  at runLifecycleMethod  &lt;src/lib/Suite.ts:351:14&gt;
  at next  &lt;src/lib/Suite.ts:525:33&gt;
  at &lt;src/lib/Suite.ts:531:15&gt;
  at runTestLifecycle  &lt;src/lib/Suite.ts:500:18&gt;
  at &lt;src/lib/Suite.ts:624:30&gt;</error>
      </testcase>
      <testcase name="constructor missing info" time="0" status="0">
        <error message="test error" type="Error">Error: test error
  at Suite.afterEach  &lt;tests/unit/lib/Environment.ts:8:11&gt;
  at &lt;src/lib/Suite.ts:383:27&gt;
  at runLifecycleMethod  &lt;src/lib/Suite.ts:351:14&gt;
  at next  &lt;src/lib/Suite.ts:525:33&gt;
  at &lt;src/lib/Suite.ts:531:15&gt;
  at runTestLifecycle  &lt;src/lib/Suite.ts:500:18&gt;
  at &lt;src/lib/Suite.ts:624:30&gt;</error>
      </testcase>
    </testsuite>
  </testsuite>
</testsuites>
```

jonnycornwell:
LGTM and I like the fact it is additive. I have not cloned/tested though as I am assume this is a draft. Once there is a final proposal/beta I am more than happy to give the changes a test. 
Not sure if you want comments on the code given its a draft but two minor things:
1. Would be nice for the `InternError` interface to have a string literal type for `lifecycleMethod` rather than string for syntactic sugar
1. JUnit reporter, lines 140-144 and 152-156 look the same so could have a shared function?

jason0x43:
It is a draft, and the points you mention do need a bit of work. Feel free to try it out if you get a chance, though. It _seems_ like it should enable the desired behavior in the JUnit reporter, although I'm not sure it's handling failures in `after` in the best way possible yet.

jonnycornwell:
@jason0x43 I just wondered what the state of this is? (we still have an internal fork sadly)

Cheers,
J

jason0x43:
I haven't had much time to work on it lately, but I'll see about cleaning it up in the next few days.

PR #958, opened 2018-10-25
Remove console reporter from defaults within the browser Html report

jonnycornwell:
This PR is to disable the console reporter default in the browser/Html reporter, in our dev environments we see a significant performance decrease with it enabled with the dev tools open (more details below).

The console reporter can still be enabled via the config if desired, e.g.

```
...
  "browser": {
    "suites": [...],
    "reporters": ["console"]
},
...
```

## Performance figures

These average were generated using our unit tests and in chrome

|         | Dev tools open       | Dev tools closed  |
| ------------- |:-------------:| -----:|
| Console On     | 132s | 54s |
| Console Off      |    83s   |   53s |

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -36,7 +36,7 @@
 				}
 
 				// Add the HTML and console reporters to the default
-				intern.configure({ 'reporters+': ['html', 'console'] });
+				intern.configure({ 'reporters+': ['html'] });
 				intern.configure(config);
 			})
 			.then(


jason0x43:
Is this significantly preferable to specifying `reporters: ['html']` in your test config? That will tell Intern to _only_ use the html reporter. I'm leery of changing defaults unless they're causing problems for many users, at least not when there's a simple alternative.

jonnycornwell:
That would be preferred, however, the reporters specified in the config are additive to the defaults. This ultimately means that you are unable to remove the console reporter through the config, unless I am missing something and there is an existing way to override defaults?

https://github.com/theintern/intern/blob/b99fe83a4816578aeb2c4cdf9f02d997e442b7c5/src/index.html#L38-L40

jason0x43:
Hmmm...I see. The browser test page should be working the same as the node runner, where there are default reporters that can be overridden.

jason0x43:
Actually, after looking into this a bit more, the existing logic works fine.

There is a top-level reporter property, and also environment-specific properties:

```json
{
  "reporters": "something",
  "browser": {
    "reporters": "html"
  },
  "node": {
    "reporters": "runner"
  }
}
```

The environment-specific reporters are merged into the top-level reporters in each environment; the assumption is that the base reporters are meant to be used everywhere, while the environment specific ones should only be used in their respective environments.

The default reporters in each environment are specified on the base property. So, in the browser, the setup looks like:

```json
{
  "reporters": ["html", "console"]
}
```

Similarly, in Node the setup looks like:

```json
{
  "reporters": ["runner"]
}
```

So if you want to override the defaults, you'll need to either set the top-level `reporters` property, or clear it and only use the environment-specific property. To remove the console reporter using a config that will only be used in the browser, you can do:

```json
{
   "reporters": "html"
}
```

To do it in a config that will be shared between Node and the browser, you would do:

```json
{
  "reporters": [],
  "browser": {
    "reporters": "html"
  },
  "node": {
    "reporters": "runner"
  }
}
```

jonnycornwell:
Thanks for the investigation and explanation, I just missed clearing the top level reporters.

Issue #950, opened 2018-09-20
Skip environments that don't match grep expression


schontz:
I have node and browser tests. In my node tests, I have a suite named `Config`. When I only want to run those tests I execute the following:

```
npx intern grep=Config
```
## Expected behavior

Only node tests should run and my browser tunnels should not load up.

## Current behavior

My browser tunnel loads up even though all the tests are skipped. There is a significant performance hit. @jason0x43  explains:
> Currently, grep is evaluated as suites are run rather than before testing starts, so Intern won't know that all tests in a suite have been filtered out before it needs to start a remote.

## Possible solution

For now, you have to manually set the environments variable:

```
npx intern environments=node grep=Config
```

jason0x43:
Probably the best way to attack this would be to evaluate grep as tests are added (i.e., don't add tests that don't match the grep expression) rather than as they're run.

schontz:
Related to this, if no tests in a suite match the grep condition, none of the before/after functions should run either.

devpaul:
Hi @schontz,

I've taken a look at this and unfortunately it's not really possible to use `grep` to determine what tests should be run before the browser tunnels are loaded. This is because `grep` works on test's ID's, which are loaded by the browser's loader. There's no universally reliable way to load these test names in node.

We did discuss one possible solution. What if you were able to filter suites by providing a part of their filename and if no suites were to run for the browser then the tunnel wouldn't start? It would look something like `intern only=Config` and only suites that had "Config" in their filename would run. By operating on filenames rather than test ids we could identify whether any suites would be sent to the browser before starting a tunnel. Would this address your use case?

Let me know what you think.

schontz:
I am not on said project anymore, but having an `only` option would address the use case I was looking at. I don't know if it would be useful for others.

jason0x43:
Closing in favor of #1022.

PR #961, opened 2018-12-02
Do not run suite lifecycle methods when all descendant tests are skipped

jonnycornwell:
- Add tests to ensure suite lifecycle functions do not run if all tests are skipped
- Evaluate grep up front rather than when the tests are executed

In our functional end-to-end tests we have had to write our own grep due to this, some of our `before` functions have substantial setup by navigating the app. Also mentioned in comment https://github.com/theintern/intern/issues/950#issuecomment-423613018.

I have gone round the houses a bit with regards to this PR and think the dust has now settled, I was getting caught out by the fact `grep` is set on the rootSuite after instantiation, in that case it needs to process all the children recursively.

With regards to https://github.com/theintern/intern/issues/950#issuecomment-423339655, I believe the tests should be added and skipped rather than just excluded.

p.s. This PR will have a merge conflict with my other PR (https://github.com/theintern/intern/pull/955) relating to the type information for `LifecycleMethod`.

### DIFF
--- a/src/lib/Suite.ts
+++ b/src/lib/Suite.ts
@@ -62,7 +62,7 @@ export default class Suite implements SuiteProperties {
   skipped: string | undefined;
 
   /** The tests or other suites managed by this suite */
-  tests: (Suite | Test)[];
+  tests: (Suite | Test)[] = [];
 
   /** The time required to run all the tests in this suite */
   timeElapsed: number | undefined;
@@ -86,8 +86,6 @@ export default class Suite implements SuiteProperties {
         const key = <keyof (SuiteOptions | RootSuiteOptions)>option;
         this[key] = options[key]!;
       });
-
-    this.tests = [];
 
     if (options.tests) {
       options.tests.forEach(suiteOrTest => this.add(suiteOrTest));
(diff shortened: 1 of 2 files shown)


jonnycornwell:
Just thought of a slight edge case related to my changes and have fixed accordingly

jonnycornwell:
Upon investigating https://github.com/theintern/intern/pull/962, I noticed that there was a race condition in the name being set after the grep following the other changes in this PR.

I have now added a failing tests for this and also a fix in 3a00d61

https://github.com/theintern/intern/blob/6a93567201e0e6cb8f9e813b5f6eaab8b4886962/src/lib/executors/Executor.ts#L798-L799.

I am 99% sure I have no more changes or PRs to raise and no more changes to make to this PR!

jason0x43:
Ooo, this will be handy. I'll try to get to this in the relatively near future.

jonnycornwell:
I am glad I left myself 1% wiggle room in my confidence. I have rolled this PR out throughout our builds and have found an issue with running unit tests via `RemoteSuite`. I haven't come across any issues when running functionalSuites or tests in the browser with the Html reporter.

Just trying to find the appropriate fix and will add it to this PR

PR #962, opened 2018-12-03
Fix query parsing when clicking on Suites/Tests in the Html reporter

jonnycornwell:
Edit: I didn't really add much context to this, in summary clicking on a link to apply a grep, results in skipping all the tests because nothing ever matches (tested in chrome 70)

The Html reporter was updated to use `URLSearchParams` at which point it started using "+" rather than "%20" to replace spaces. The knock on effect was that `parseQuery` does not understand what to do with "+" as it uses `decodeURIComponent`.

https://github.com/theintern/intern/blob/6a93567201e0e6cb8f9e813b5f6eaab8b4886962/src/lib/browser/util.ts#L121

I did not parse using `URLSearchParams` as I presume tests are expected to [run in IE](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams#Browser_compatibility). I am not sure if this was intentional for the Html reporter but tests using `Runner` should continue to work in IE.

### DIFF
--- a/src/lib/browser/util.ts
+++ b/src/lib/browser/util.ts
@@ -108,22 +108,15 @@ export function normalizePath(path: string) {
  * Parse a query string and return a set of decoded name=value pairs
  */
 export function parseQuery(query?: string) {
-  query = query || global.location.search.slice(1);
-  return query!
-    .split('&')
-    .filter(arg => {
-      return arg !== '' && arg[0] !== '=';
-    })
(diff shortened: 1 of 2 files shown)


jason0x43:
Good catch. The browser bundle uses a polyfill for URLSearchParms, so we should be able to use that for parsing.

jonnycornwell:
Ok, will update accordingly

PR #968, opened 2019-02-07
Fix expand bug in html reporter that causes a row to disappear.

### LINKED ISSUE #966: Opening collapsed sub section on web view hides below child
## Expected behavior

The next top level section should be pushed downward and not disappear

## Current behavior
Opening collapsed sub section on web view hides below child 
http://g.recordit.co/T4NwW430Y3.gif

## Steps to reproduce (for bugs)

1. Create two top level sections, the first one should have at least one sub section
2. open the first top level section
3. open the subsection within the first top level section

the second top level section has disappeared

## Environment

Intern version: 4.2.0
Node version: 8.12.0
NPM version: 6.4.1
Browser version: Chrome Version 70.0.3538.110 (Official Build) (64-bit)

edhager:
Fixes #966.

When a row is expanded in the HTML reporter, only direct descendants are made visible all grandchildren are hidden.  The code processing the rows was written to stop processing rows if a row at the same indent level was found.  If a row with a smaller indent value was encountered, that row was hidden as if it were a descendant.

### DIFF
--- a/src/lib/reporters/Html.ts
+++ b/src/lib/reporters/Html.ts
@@ -195,7 +195,7 @@ export default class Html extends Reporter implements HtmlProperties {
 
       // Stop looping when we encounter a row that's not indented more
       // than the suite being updated
-      if (indentDelta === 0) {
+      if (indentDelta <= 0) {
         break;
       }
 


jason0x43:
This appears to do the trick. Thanks!

PR #969, opened 2019-02-14
Make file URLs in error output more user friendly.

### LINKED ISSUE #952: URL links in console do not work (404)
When running `intern serveOnly` and a test fails there is an error in the console.

## Expected behavior
Clicking on the URL in the error should load the file in Sources.

## Current behavior
Clicking on the URL in the console opens a 404 in a new tab.

Sample URL: http://localhost:9000/__intern/e.Should%20support%20error()%20logging%20[as%20test]%20%20%3Chttp:/localhost:9000/tests/unit/loggerBrowser.js:15:36%3E

## Environment

Intern version: 4.3.1
Browser version: Chrome on Windows

edhager:
Fixes #952.

This PR makes URLs contained in errors printed to the browser console more user friendly by doing two things:

1. Spaces are added around the URL so the browsers do not include the ">" as part of the URL:
   ```
   at regular suite  < http://localhost:9000/_tests/tests/unit/lib/reporters/Html.js:87:21 >
   ```
   With this change, clinking on the link does the following:

   - Chrome opens the file at the designated line number in the "Sources" tab in the development tools.
   - Safari, opens the file at the designated line number in the "Resources" tab in the development tools.
   - Firefox tries to open the file in a new tab.  Without the next change, this request fails.
   - IE and Edge do nothing because they do not turn the URL into a hyperlink.  
1. To support the request that Firefox makes when the URL is clicked in the console, a URL filter was added to the server code to remove the line and column number references so the request is handled properly.


jason0x43:
Finally got a chance to look at this. In my own tests, it seems to work as described -- great!

I agree, though, that the spaces seem less than ideal. What if we used a format more like Chrome's for the locations, like `at regular suite @ http://localhost:9000/_tests/tests/unit/lib/reporters/Html.js:87:21`? The `<>` were originally used so the output would look similar to browser stack listings. However, browsers seem to have adopted simpler formats at this point.

edhager:
@jason0x43 I updated the code to use `@` instead of `< >`.

PR #978, opened 2019-03-12
Expose requested environment on Remote

ascorbic:
Different drivers are inconsistent in how they report the browser and platform name and version, and IE for example doesn't report the OS version at all. This makes it hard to check the current platform from within tests. This PR adds a `requestedEnvironment` property to the `remote` object so that you can see the environment that you requested.


jason0x43:
There's still a chance that the OS or browser version you end up with might not be exactly what you asked for, but I can still see this being a useful indicator.

ascorbic:
That's great. Thanks for merging. When can we expect the next release?

PR #995, opened 2019-09-23
Allows running and providing coverage for TS tests and code

### LINKED ISSUE #973: Support transparent loading of TS modules
Intern should be able to use transparently load TS modules for Node-based tests, either by default or via a plugin. Intern 4 can currently be used with ts-node at a basic level. However, doing so requires a small amount of custom code, and coverage data isn't recorded for modules loaded via ts-node.

Ideally, a user should be able to provide TS modules as resources (`suites`, `functionalSuites`, `plugins`) and just have that work.

maier49:
Registers ts-node before running tests, optionally with a custom path to a `tsconfig.json` file if the project's should not be picked up automatically. Also extends the istanbul coverage hook config to provide coverage for these files.

Resolves #973


src/lib/common/config.ts
```diff
@@ -476,6 +476,11 @@ export interface ResourceConfig {
    */
   suites: string[];
 
+  /**
+   * An optional path to a tsconfig for ts-node. This options can be used
```
jason0x43:
```suggestion
   * An optional path to a tsconfig for ts-node. This option can be used
```

src/lib/executors/Node.ts
```diff
@@ -631,6 +644,10 @@ export default class Node extends Executor<NodeEvents, Config, NodePlugins> {
         }
       }
 
+      this.config.node.tsconfig
```
jason0x43: I think an `if/else` may be clearer for imperative code. Also, we should provide some way to disable ts-node/register, like by setting `config.node.tsconfig` to `false` or something, in case someone doesn't want it loaded.

src/lib/executors/Node.ts
```diff
@@ -715,7 +732,7 @@ export default class Node extends Executor<NodeEvents, Config, NodePlugins> {
       // handle suites
       this._instrumenter = createInstrumenter(
         Object.assign(
-          {},
+          { esModules: true },
```
jason0x43: This should probably be optional, based on whether ts-node is in use.

src/lib/executors/Node.ts
```diff
@@ -886,7 +903,8 @@ export default class Node extends Executor<NodeEvents, Config, NodePlugins> {
     );
     this._unhookRequire = hookRequire(
       filename => this.shouldInstrumentFile(filename),
-      (code, { filename }) => this.instrumentCode(code, filename)
+      (code, { filename }) => this.instrumentCode(code, filename),
+      { extensions: ['.js', '.jsx', '.ts', 'tsx'] }
```
jason0x43: What about `mjs`?

src/lib/node/util.ts
```diff
@@ -13,6 +13,7 @@ import {
   parseArgs,
   splitConfigPath
 } from '../common/util';
+import * as path from 'path';
```
jason0x43: Why not add `extname` to the list of path imports higher up?

tests/unit/lib/middleware/instrument.ts
```diff
@@ -22,7 +22,7 @@ registerSuite('lib/middleware/instrument', function() {
 
   let server: Server;
   let shouldInstrumentFile: sinon.SinonStub<[string]>;
-  let instrumentCode: sinon.SinonStub<[string, string]>;
+  let instrumentCode: sinon.SinonStub<[string, string, (boolean | undefined)?]>;
```
jason0x43: Isn't `(boolean | undefined)?` a bit redundant?

src/lib/executors/Node.ts
maier49: 👍

src/lib/common/config.ts
maier49: Fixed the typo and added a note about setting it to `false` to avoid registering ts-node at all

src/lib/executors/Node.ts
maier49: The default was only `.js` (or at least that's what I gathered looking at the code, didn't see anywhere it was documented).  Even after registering tsnode `.mjs` doesn't exist in `require.extensions` so I'm not sure we should or even could add that.

tests/unit/lib/middleware/instrument.ts
maier49: Yes it is. I just used what sinon spit out at me in the error but I shortened it to `boolean?`

src/lib/executors/Node.ts
jason0x43: Thanks for looking into that. We can always get back to it later if it's an issue.

src/bin/intern.ts
```diff
@@ -25,6 +25,18 @@ getConfig()
 
       intern.configure({ reporters: 'runner' });
       intern.configure(config);
+
+      if (
+        (intern.environment === 'browser' &&
+          (intern.config.suites &&
+            intern.config.suites.some(pattern => pattern.endsWith('.ts')))) ||
+        (intern.config.plugins &&
```
jason0x43: I think a paren may be off here. This is saying `if (browser || plugins) { ... }`, but plugins in node should be OK.

src/lib/executors/Node.ts
```diff
@@ -757,7 +776,11 @@ export default class Node extends Executor<NodeEvents, Config, NodePlugins> {
       // handle suites
       this._instrumenter = createInstrumenter(
         Object.assign(
-          {},
+          {
+            esModules: Boolean(
+              typeof config.tsconfig === 'undefined' || config.tsconfig
```
jason0x43: Should this depend on the `compilerOptions.module` property in the tsconfig? It shouldn't be necessary if, for example, the module style is "commonjs".

src/lib/executors/Node.ts
```diff
@@ -876,7 +899,7 @@ export default class Node extends Executor<NodeEvents, Config, NodePlugins> {
       uncoveredFiles.forEach(filename => {
         try {
           const code = readFileSync(filename, { encoding: 'utf8' });
-          this.instrumentCode(code, filename);
+          this.instrumentCode(code, filename, true);
```
jason0x43: Should this always be true, or should it depend on the extension?

maier49: It should (and now does) depend on the extension

src/lib/executors/Node.ts
maier49: I actually think right now this is not necessary...if the user wants to use es6 modules they can pass the instrumenterConfig themselves, and it's quite a pain to figure out what the `module` type would actually be from here, especially considering tsconfig extension.

src/bin/intern.ts
maier49: Whoops, added more test cases to better verify this behavior and fixed those parens.

PR #1003, opened 2019-12-11
Update tslib to 1.10

devpaul:
VSCode was producing a warning when using an array spread:

image

The `__spreadArrays` method was added to tslib 1.10 and used by TS as an improvement on array spreads w/o downlevel iteration. This applies to Intern. (though we may be getting lucky as the compiler doesn't use `__spreadArrays` [for simple instances](https://github.com/microsoft/TypeScript/pull/31166#pullrequestreview-247774327)).


devpaul:
This warning is due to VSCode using it's own internal version of TS (3.6) rather than the workspace's version by default. After switching to use the workspace's TS the error goes away!

We can address this later when we update TS.

PR #1005, opened 2019-12-12
Retry sessions

### LINKED ISSUE #996: Implement session-level retry
When performing a test run with multiple browser sessions, such as Intern's own CI tests, one or two browsers may sporadically fail while the others succeed, leading to a failure of the overall test run. In that situation (where only some browser sessions fail), Intern could automatically re-run the failing browser sessions at least once in case the failure was a fluke.

This is similar to #340, but at the session level rather than the individual test level.

devpaul:
Resolves #996

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -114,6 +114,7 @@ Some properties are only meaningful for Node or WebDriver tests:
 | [capabilities]       | Default capabilities to be used for WebDriver sessions                                                                                                           | `{ 'idle-timeout': 60 }`    |
 | [coverage]           | An array of paths or globs to collect coverage data for                                                                                                          | `[]`                        |
 | [environments]       | Browser + OS combinations to be tested using WebDriver                                                                                                           | `[]`                        |
+| [functionalRetries]  | rerun failed sessions when at least one other session succeeds                                                                                                   | `0`                         |
 | [functionalSuites]   | Suites to run in WebDriver mode                                                                                                                                  | `[]`                        |
 | [functionalTimeouts] | Timeouts used in functional tests                                                                                                                                | `{ connectTimeout: 30000 }` |
 | [leaveRemoteOpen]    | If true, leave remote browsers open after testing has finished                                                                                                   | `false`                     |
@@ -561,6 +562,8 @@ There are a few exceptions:
 [extends]: #extends
 [filtererrorstack]:
   https://theintern.io/docs.html#Intern/4/api/lib%2Fcommon%2Fconfig/filtererrorstack
+[functionalretries]:
+  https://theintern.io/docs.html#Intern/4/api/lib%2Fcommon%2Fconfig/functionalretries
 [functionalsuites]:
   https://theintern.io/docs.html#Intern/4/api/lib%2Fcommon%2Fconfig/functionalsuites
 [functionaltimeouts]:
--- a/schemas/config.json
+++ b/schemas/config.json
@@ -248,6 +248,11 @@
           "type": "boolean"
         },
 
+        "functionalRetries": {
+          "description": "The number of times to rerun failed sessions when at least one other session succeeds",
+          "type": "number"
+        },
+
         "functionalSuites": {
           "description": "Suites to run as functional (WebDriver) tests",
           "$ref": "#/definitions/suitesProperty"
--- a/src/core/lib/Suite.ts
+++ b/src/core/lib/Suite.ts
@@ -295,7 +295,6 @@ export default class Suite implements SuiteProperties {
    * Add a test or suite to this suite.
    */
   add(suiteOrTest: Suite | Test) {
-    console.log(suiteOrTest);
     if (!isTest(suiteOrTest) && !isSuite(suiteOrTest)) {
       throw new Error('Tried to add invalid suite or test');
     }
@@ -351,7 +350,6 @@ export default class Suite implements SuiteProperties {
    */
   reset() {
     this._remote = undefined;
-    // Remove implicit reset from run() (in case it's relied upon)
     this.error = undefined;
     this.timeElapsed = 0;
   }
@@ -520,12 +518,6 @@ export default class Suite implements SuiteProperties {
     const after = () => {
       return runLifecycleMethod(this, 'after');
     };
-
-    // Reset some state in case someone tries to re-run the same suite
-    // TODO: Cancel any previous outstanding suite run
-    // TODO: Test
-    this.error = undefined;
-    this.timeElapsed = 0;
 
     let task: CancellablePromise<void>;
     let runTask: CancellablePromise<void>;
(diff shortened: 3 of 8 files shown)


src/lib/Suite.ts
```diff
@@ -213,6 +213,15 @@ export default class Suite implements SuiteProperties {
...
   }
 
+  get failed(): boolean {
+    return this.tests.some(suiteOrTest => {
+      return (
+        !!this.error ||
+        (isSuite(suiteOrTest) ? suiteOrTest.failed : suiteOrTest.error)
+      );
+    });
+  }
```
jason0x43: Could this just be `return this.numFailedTests > 0 || this.error`?

src/lib/common/config.ts
```diff
@@ -199,6 +199,9 @@ export interface Config extends ResourceConfig {
   /** Whether to collect coverage data from functional tests */
   functionalCoverage: boolean;
 
+  /** The number of times to rerun a session when a session fails when another one succeeds */
```
jason0x43: Maybe "when it fails and at least one other session succeeds"

src/lib/executors/Node.ts
```diff
@@ -875,13 +876,34 @@ export default class Node extends Executor<NodeEvents, Config, NodePlugins> {
...
-              () => (testTask = this._runRemoteTests())
-            );
+
+            let remainingAttempts = 1 + (this.config.functionalRetries || 0);
+            let suites = this._sessionSuites;
+
+            await this._loadFunctionalSuites();
+            while (remainingAttempts && suites.length) {
+              remainingAttempts--;
+              if (suites.length !== this._sessionSuites!.length) {
```
jason0x43: It seems odd that `this._sessionSuites` needs a `!` but `suites` doesn't.  Consider adding `const sessionSuites = this._sessionSuites!` and using that to avoid having a bunch of non-null assertions.

src/lib/executors/Node.ts
```diff
@@ -875,13 +876,34 @@ export default class Node extends Executor<NodeEvents, Config, NodePlugins> {
...
+                this.log(
+                  'reattempting',
+                  suites.length,
+                  'of',
+                  this._sessionSuites!.length,
+                  'environments'
+                );
+              }
+              testTask = this._runRemoteTests(suites);
+              await testTask.catch(remainingAttempts ? () => {} : undefined);
```
jason0x43: Why not a `try/catch`?

src/lib/executors/Node.ts
```diff
@@ -875,13 +876,34 @@ export default class Node extends Executor<NodeEvents, Config, NodePlugins> {
...
+              suites = suites.filter(suite => suite.failed);
```
jason0x43: Consider using a meaningful name here (maybe `failingSuites`), just to make it clearer what's being done in the following if statement, then assign that to `suites` at the end of the loop.

jason0x43 (changes requested):
This looks good overall. I just had a few cleanup suggestions, and the `Suite.failed` method could probably be simplified.

src/lib/executors/Node.ts
devpaul: This seemed less verbose than

src/lib/executors/Node.ts
devpaul: It doesn't need the `!` any longer. Removed.

src/lib/executors/Node.ts
devpaul: Hmm... The conditional check is on the next line. Do you think it needs an explicit name when usage is so close to assignment?

jason0x43: It makes it read better. It was the usage specifically that prompted the comment; it's not immediately apparent when skimming through the code that `if (suites.length === this._sessionSuites.length)` is comparing a list of failing suites to the original list.

src/lib/executors/Node.ts
jason0x43: I think we should be more explicit in general; `catch(remainingAttempts ? () => {} : undefined)` obscures the fact that the goal is to consume or throw the error. If you expand the catch, it's not much different than the `try / catch`, so we should probably use that for consistency.

jason0x43:
The `config.json` schema and the docs should also be updated for the new config parameter.

src/lib/Suite.ts
```diff
@@ -346,6 +346,16 @@ export default class Suite implements SuiteProperties {
     );
   }
 
+  /**
+   * Explicity reset the suite so it may run again
+   */
+  reset() {
+    this._remote = undefined;
+    // Remove implicit reset from run() (in case it's relied upon)
```
jason0x43: Since this is going in v5, go ahead and actually remove the implicit reset code from Suite. Move the TODO about canceling suite runs here.

src/lib/executors/Node.ts
```diff
@@ -875,13 +876,46 @@ export default class Node extends Executor<NodeEvents, Config, NodePlugins> {
...
+            const allSessions = this._sessionSuites;
+            let suites = allSessions;
+            let remainingAttempts = 1 + (this.config.functionalRetries || 0);
+
+            await this._loadFunctionalSuites();
+            while (remainingAttempts && suites.length) {
+              remainingAttempts--;
+              if (suites.length !== allSessions.length) {
+                this.log(
+                  'reattempting',
```
jason0x43: We ideally shouldn't say 'reattempting' on the first run.

src/lib/executors/Node.ts
```diff
@@ -875,13 +876,46 @@ export default class Node extends Executor<NodeEvents, Config, NodePlugins> {
...
+                );
+              }
+              try {
+                testTask = this._runRemoteTests(suites);
+                await testTask;
+              } catch (e) {
+                this.log(`suite error: ${e}`);
+                // recover from exceptions to allow for retries
+              }
+              const failedSuites = (suites = suites.filter(suite =>
```
jason0x43: The filter statement can just be `suites.filter(isFailedSuite)`

src/lib/executors/Node.ts
```diff
@@ -875,13 +876,46 @@ export default class Node extends Executor<NodeEvents, Config, NodePlugins> {
...
+          .then(async () => {
             if (!this._sessionSuites) {
               return;
             }
-            return this._loadFunctionalSuites().then(
-              () => (testTask = this._runRemoteTests())
-            );
+
+            const allSessions = this._sessionSuites;
+            let suites = allSessions;
```
jason0x43: Could we call this `sessions`, and have `failedSessions`? The naming scheme is a bit inconsistent between here and `_runRemoteTests`.

jason0x43 (changes requested):
Just a couple more changes (it was going to need some rework before the merge, anyway 😄)

src/lib/executors/Node.ts
devpaul: suites.length and allSessions.length will be equal on the first run. Suites are filtered later as successful ones are removed.

jason0x43: Ah, good point

jason0x43 (approved):

PR #1008, opened 2019-12-17
Marshal suiteError and centralize JSON error marshalling

jonnycornwell:
Looks like I missed this with my original testing, sorry it is going in as a separate PR!

The suiteError is not getting marshalled to JSON in `Test.toJSON()` and therefore the JUnit reporter is not working as expected, see before and after screenshot below.

image

This time I have centralised "common" error marshalling to remove the duplication between the `Suite` and the `Test`.  Wasn't really sure where to put the function as I didn't really want to make a new file so utils seemed like the best location given the existing file structure. 

On the side I am looking at a slightly larger refactor around the JSON marshalling and the `RemoteSuite` as the life cycle events are somewhat incorrect with regards to the typings. For example, the `suiteEnd` callback claims to be given a `Suite` but it is actually the marshalled JSON or a `Suite`, therefore something like `Suite.remote` will never be defined in a reporter whilst running unit tests in the browser (it will be set on the root of the `RemoteSuite`).

### DIFF
--- a/src/lib/common/util.ts
+++ b/src/lib/common/util.ts
@@ -4,6 +4,7 @@ import { Config, ResourceConfig } from './config';
 import { Events, Executor, PluginDescriptor } from '../executors/Executor';
 import { TextLoader } from './util';
 import { getPathSep, join, normalize } from './path';
+import { InternError } from '../types';
 
 export interface EvaluatedProperty {
   name: keyof Config;
@@ -996,4 +997,28 @@ function serializeReplacer(_key: string, value: any) {
   }
 
   return value;
+}
+
+export function errorToJSON(error?: InternError): InternError | undefined {
+  if (!error) {
+    return undefined;
+  }
+  const {
+    name,
+    message,
+    stack,
+    lifecycleMethod,
+    showDiff,
+    actual,
+    expected
+  } = error;
+
+  return {
+    name,
+    message,
+    stack,
+    ...(lifecycleMethod ? { lifecycleMethod } : {}),
+    showDiff: Boolean(showDiff),
+    ...(showDiff ? { actual, expected } : {})
+  };
 }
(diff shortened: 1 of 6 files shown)


src/lib/common/util.ts
```diff
@@ -997,3 +998,19 @@ function serializeReplacer(_key: string, value: any) {
 
   return value;
 }
+
+export function marshalErrorToJSON(
```
jason0x43: Maybe just `errorToJson`

src/lib/common/util.ts
```diff
@@ -997,3 +998,19 @@ function serializeReplacer(_key: string, value: any) {
...
+  if (!error) {
+    return undefined;
+  }
+  const { name, message, stack, lifecycleMethod } = error;
+
+  return {
+    name,
+    message,
+    stack,
+    ...(lifecycleMethod ? { lifecycleMethod } : {})
```
jason0x43: For consistency, this method should handle other InternError fields as well, or at least those we currently care about (e.g., `showDiff`).

jason0x43:
> For example, the suiteEnd callback claims to be given a Suite but it is actually the marshalled JSON or a Suite, therefore something like Suite.remote will never be defined in a reporter whilst running unit tests in the browser (it will be set on the root of the RemoteSuite).

Do you mean it's not accessible in the browser, or that it's not accessible to a reporter running on the Intern host?

jonnycornwell:
> Do you mean it's not accessible in the browser, or that it's not accessible to a reporter running on the Intern host?

Sorry, that was poorly worded! So the issue is on the node.js side when using the RemoteSuite, the actual `Suite`/`Test` objects passed into the reporters are the results of the toJSON function rather than the `Suite`/`Test` classes as the API suggests.

Just applying the changes you've suggested now :)

PR #1010, opened 2019-12-24
Option to disable the 'dom' reporter in the remote browser.

### LINKED ISSUE #909: Provide an equivalent to for Intern 3's writeHtml option
The `writeHtml` option was used to prevent the runner reporter from writing to the DOM during tests in cases where unit tests might be modifying the page (#435). The Dom reporter in Intern 4 has the same concerns, and it should be made more resilient against DOM modifications and/or Intern should provide an option to disable it.

rhpijnacker:
As discussed on Gitter, this creates an option
`{
    remoteOptions: {
        disableDomUpdates: true
    }
}`
that avoids the `dom` reporter from being automatically selected for a remotely opened browser run.

This fixes #909.

### DIFF
--- a/src/lib/RemoteSuite.ts
+++ b/src/lib/RemoteSuite.ts
@@ -221,11 +221,14 @@ export default class RemoteSuite extends Suite {
         // These are options that will be POSTed to the remote page and
         // used to configure intern. Stringify and parse them to ensure
         // that the config can be properly transmitted.
+        const disableDomUpdates =
+          config.remoteOptions && config.remoteOptions.disableDomUpdates;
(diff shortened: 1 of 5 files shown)


jason0x43:
Could you also add a parser for the new option to `src/lib/common/util.ts`, and a self test for the new property in `tests/uni/lib/executors/Executor.ts` (in the `known properties` subsuite)?

rhpijnacker:
Added the parser and testcase. Is this what you had in mind?
By the way, it took me some time to figure out that running the self test with `npm run test` does not work properly on Windows. Are you aware of this?

jason0x43:
The parser and test case look good.

I believe someone might have recently mentioned a Windows issue; I'll look into it.

PR #1012, opened 2020-01-02
make listerner return PromiseLike interface

### LINKED ISSUE #949: Typing issue for event listeners that return PromiseLikes
In 4.2.x and 4.3.x, the Listener interface return type is void or a Promise. That makes TS too picky about what is returned, resulting in typing errors when promise-like objects (e.g., Leadfoot Commands) are returned from handlers. It should instead look for a PromiseLike.

aciccarello:
closes #949


aciccarello:
Still need to confirm that this fixes the issue described in #949

PR #1014, opened 2020-01-09
Added directory/filename support to JUnit

### LINKED ISSUE #985: Make options consistent for reporters that generate output
## Expected behavior

For all reporters that generate output it would be useful for them to take a consistent set of options.

## Current behavior

At the moment some reporters, such as the `lcov` and `htmlcoverage` accepts a `directory` and `filename` (in lcov's case). However the `junit` reporter only accepts a `filename` option.

## Possible solution

Normalise the options for reporters to always accept `directory` and if required a `filename`. This could be made backwards compatible in the `junit` reporters case by using the `filename` as the full path if no `directory` is provided.

## Steps to reproduce (for bugs)

N/A

## Environment

Intern version: 4.4.2
Node version: 8.9.1
NPM version: 6.4.1

devpaul:
Added `getPath`, which handles the progressive fallback of directory/filename options.

I also included comments in `Coverage` that includes my findings on how it uses `filename` and `directory` as a base class (throws them in bags and hands them off to the Istanbul reporters).

The functionality here is done, but perhaps it could use some cleanup since the work has been isolated to JUnit (due to the above discoveries)

Resolves #985

### DIFF
--- a/docs/api.json
+++ b/docs/api.json
@@ -5,7 +5,7 @@
 	"flags": {},
 	"children": [
 		{
-			"id": 4346,
+			"id": 4350,
 			"name": "\"bin/intern\"",
 			"kind": 1,
 			"kindString": "External module",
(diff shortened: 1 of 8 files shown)


devpaul:
This PR will be applied after the repo reunion: #998

jason0x43:
After considering this a bit more, I lean towards putting this one on hold until we get around to making more global updates to the reporter system and to Intern's configuration. The root problem is that users need a standard way to specify options for file based reporters, and it seems better to address that all at once rather than piecemeal.

In the end we should have a couple of interfaces (e.g., FileReporterOptions, PathReporterOptions) to cover reporters that output files (like JUnit) and reporters that output into directories (like HtmlCoverage). Reporter options should inherit from one of these (or neither if they don't output files), and the available options for each built in reporter should be encoded in the config schema.

devpaul:
Closed. Notes added to #1023

PR #1017, opened 2020-01-10
Draft: Force enabling scrolling in html reporter

rhpijnacker:


jason0x43:
I mentioned this in gitter; just repeating it here.

Would it make more sense to have the HTML reporter disable any stylesheets from the document rather than trying to catch any global styles that might be hanging around? Maybe something like this in `reporters/Html#suiteEnd`:

```ts
if (!suite.hasParent) {
  this._generateSummary(suite);
  
  for (let i = 0; i < document.styleSheets.length; i++) {
    document.styleSheets[i].disabled = true;
  }
  
  require('./html/html.styl');
  // ...
}
```

rhpijnacker:
Yes, that makes perfect sense since this is a much cleaner solution.

PR #1025, opened 2020-01-14
Disable existing stylesheets in Html reporter

rhpijnacker:
The Html reporter needs some css rules to properly show the overview. 
The css rules of the executed test cases also add styling to the browser, which can interfere with the way the overview is rendered. E.g. this can cause scrolling to be disabled.
This PR disables the existing stylesheets to make sure that only the proper css rules are applied.


rhpijnacker:
Somehow this PR ended up being a mess. Opened 1026 to fix this.

PR #1026, opened 2020-01-14
Disable existing stylesheets in Html reporter

rhpijnacker:
The Html reporter needs some css rules to properly show the overview.
The css rules of the executed test cases also add styling to the browser, which can interfere with the way the overview is rendered. E.g. this can cause scrolling to be disabled.
This PR disables the existing stylesheets to make sure that only the proper css rules are applied.

### DIFF
--- a/src/lib/reporters/Html.ts
+++ b/src/lib/reporters/Html.ts
@@ -473,6 +473,10 @@ export default class Html extends Reporter implements HtmlProperties {
     if (!suite.hasParent) {
       this._generateSummary(suite);
 
+      for (let i = 0; i < document.styleSheets.length; i++) {
+        document.styleSheets[i].disabled = true;
+      }
+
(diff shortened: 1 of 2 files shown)


jason0x43:
This looks good. Could you also add a unit test to the 'root suite' test in `tests/unit/lib/reporters/Html.ts` to verify that the reporter is disabling stylesheets?

jason0x43:
Actually, adding a test for that may be non-trivial given how the mock document currently works. If it looks like it'll be high-effort, just add a `// TODO:` comment in the 'root suite' test saying to add a unit test for stylesheet disabling.

rhpijnacker:
I tried to add a simple test that adds a `<style type="text/css">div { .width: 100%; }<style>`, but when asking the (mock) document to its styleSheets it just always returns 0.

PR #1053, opened 2020-01-19
Serialize events

rhpijnacker:
Individual calls to `emit(<event>)` currently start a new asynchronous promise chain. In the [intern-cucumber](https://github.com/rhpijnacker/intern-cucumber) plugin, this causes overlapping handling of the events, since the `suiteEnd` handling is not yet finished before the new `suiteStart` event is triggered.
This problem is easily fixed by chaining this `suiteStart` event at the end of the `suiteEnd` chain.


rhpijnacker:
PR to the wrong branch... :o ... sorry

PR #1090, opened 2020-02-04
feat(bdd): add xit and xdescribe methods to bdd interface

### LINKED ISSUE #595: xit and xdescribe methods in bdd interface
I find myself incrementally testing methods and commenting out some without having them run or report. I was using `this.skip` but have been wanting to block out bigger chunks as well. At this point i've just been commenting out blocks of code in my test files which at some points becomes a big cumbersome.

Are there any issues with providing an interface for `bdd` for `xit` and `xdescribe` methods with a possible noop method or at the least report based on description of block that is being bypassed?

I think this would be more convenient then the `this.skip()` method since I can deactivate entire describe block. If it's approved, I can just update the interface it looks like here with a noop method for both `xit` and `xdescribe`:

https://github.com/theintern/intern/blob/2c209d96820020080f26efd84fab510510afeaab/lib/interfaces/bdd.js

```
return Object.create(tdd, {
describe: createValue(tdd.suite),
it: createValue(tdd.test),
suite: { value: undefined },
test: { value: undefined },
xdescribe: function() {},
xit: function() {}
});
```

aciccarello:
test structure is taken from tdd tests.

closes #595

### DIFF
--- a/src/core/lib/interfaces/bdd.ts
+++ b/src/core/lib/interfaces/bdd.ts
@@ -5,7 +5,9 @@ import {
   TddSuiteFactory,
   getInterface as getTddInterface,
   suite,
+  xsuite,
   test,
+  xtest,
   before,
   after,
   beforeEach,
@@ -14,21 +16,34 @@ import {
 
 export interface BddInterface extends TddLifecycleInterface {
   describe(name: string, factory: TddSuiteFactory): void;
+  xdescribe(name: string, factory: TddSuiteFactory): void;
   it(name: string, test: TestProperties['test']): void;
+  xit(name: string, test?: TestProperties['test']): void;
 }
 
 export { suite as describe };
+export { xsuite as xdescribe };
 export { test as it };
+export { xtest as xit };
 export { before, after, beforeEach, afterEach };
 
 export function getInterface(executor: Executor): BddInterface {
-  const { suite, test, before, after, beforeEach, afterEach } = getTddInterface(
-    executor
-  );
+  const {
+    suite,
+    xsuite,
+    test,
+    xtest,
+    before,
+    after,
+    beforeEach,
+    afterEach
+  } = getTddInterface(executor);
 
   return {
     describe: suite,
+    xdescribe: xsuite,
     it: test,
+    xit: xtest,
     before,
     after,
     beforeEach,
--- a/src/core/lib/interfaces/tdd.ts
+++ b/src/core/lib/interfaces/tdd.ts
@@ -34,7 +34,9 @@ import { Executor } from '../executors/Executor';
 
 export interface TddInterface extends TddLifecycleInterface {
   suite(name: string, factory: TddSuiteFactory): void;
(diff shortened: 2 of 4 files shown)


src/core/lib/interfaces/bdd.ts
```diff
@@ -14,21 +14,40 @@ import {
...
 }
 
 export { suite as describe };
 export { test as it };
 export { before, after, beforeEach, afterEach };
 
+export function xdescribe(name: string, factory: TddSuiteFactory) {
+  return suite(name, function(suiteRef) {
+    before(function() {
+      suiteRef.skip();
```
jason0x43: This should provide a skip reason to give the user some indication of why the suite was skipped ("xdescribe", "not implemented", ...)).

src/core/lib/interfaces/bdd.ts
```diff
@@ -14,21 +14,40 @@ import {
...
+    });
+    factory(suiteRef);
+  });
+}
+
+export function xit(name: string, _test?: TestProperties['test']) {
+  return test(name, function() {
+    this.skip();
```
jason0x43: This should provide a skip reason to give the user some indication of why the suite was skipped ("xit", "not implemented", ...).

aciccarello:
@jason0x43 For the case when `xit` is called without a function we could mark the test as "xit: not implemented". For the other cases, I think "xdescribe" and "xit" are probably better descriptions because they could be skipped for any number of reasons.

If we want we can add the xsuite and xtest interfaces however I've never seen a tdd interface.

RoystonS:
Btw, a quick API idea that we put into our local `xit` and `xdescribe` methods is to add an optional trailing 'reason' parameter to allow the recording of the skip reason.

e.g.
```
it('rubs the lotion on its skin', () => {
  // test code
});
```
becomes
```
xit('rubs the lotion on its skin', () => {
  // test code
}, 'not working properly after build machine Firefox upgrade');
```
We've found that more useful than a code comment because the skip reason can then appear in the test output.

jason0x43:
@aciccarello Your thoughts on the comments sound fine to me. @RoystonS's idea of adding an optional third parameter for the `x` functions also sounds pretty reasonable -- we could default to the `x` skip reason, but allow the user to give something more meaningful if they wanted to.

As far as the tdd interface goes, it's just a clone of bdd with more explicit function names. (Well, it's actually the other way around.) They're intended to be functionally equivalent, so if we add something to one, we should add it to the other.

jason0x43:
Given their similarity, though, we should probably just drop the tdd interface entirely and let users import bdd with different names if they want to.

```
import { describe as suite, it as test } from 'intern/core/lib/interfaces/bdd';
```

jason0x43:
After more discussion, it's probably best to move the new functionality to the tdd interface (with everything else) and just import and export it from bdd.

aciccarello:
@jason0x43 Updated to include xsuite and xtest

PR #1091, opened 2020-02-04
Switch to standard Promises

### LINKED ISSUE #1018: Switch to standard Promises
Intern 4.x makes extensive use of a Task class that provides cancellation and `finally`. Now that `finally` is part of standard promises, and given that cancellation isn't all that useful, the Task class should be replaced with Promise, and all APIs returning Tasks should now return PromiseLike.

The webdriver Command class should remain a custom Promise, but based on Promise rather than Task / CancellablePromise.

jason0x43:
This PR will remove the custom Task and CancellablePromise classes in favor of standard Promises.

Along with removing Task, this PR also aims to replace task cancellation with some equivalent functionality. My current plan is cancellation tokens, which should be easy enough from the user perspective; a user should only need to deal with tokens at a few calls (`intern.run`, the start of a Command chain). We could possibly forgo most explicit token handling for the user by using  `cancel` methods on Executor and Command. In any case, most of the work will be involved in propagating the token around Intern's internals.

resolves #1018


jason0x43:
When a user cancels an operation, Intern should stop long running operations, if possible. That means we can't just wrap high-level promises, because even though canceling a cancellation token would short circuit an Intern run, it wouldn't cause the underlying long operation to stop.

```ts
this.runPromise = token.wrap(this._runTests())
```

Cancelling the token would make `runPromise` reject immediately, but it wouldn't cause the tests started by `_runTests` to stop running.

Cancellation should be applied at as low a level as is reasonable. Token values should also be checked before starting any significant synchronous work. The goal is for the testing process to end relatively quickly after a user calls `cancel`, not necessarily immediately.

jason0x43:
Cancellation with Promises will work differently than it did with Tasks. Cancelling a Task was a synchronous operation that would immediately execute cancellation logic. Cancellation in this PR is based around Promise rejection. When a Suite/Test/Executor is cancelled, its run operation will reject with a cancellation error; this error can be identified as a cancellation using an `isCancel` utility function.

jason0x43:
In general, cancellation should be handled at the lowest level and allowed to down any promise chains. That will allow internal tasks time to clean up before the cancellation is reported for an overall task.

For example, in executor's `run` method, one option would be to wrap the top-level with a cancel token. If the token is canceled, the `run` promise would reject immediately, and any sub-promises wrapped by the token would also be canceled. However, the code that `called` the run method wouldn't know anything about child tasks. As soon as the `run` promise resolved, it might decide to call `run` again, potentially before internal tasks had finished.

PR #1100, opened 2020-02-14
feat(1035): check for tsconfig and functionSuites ts file

### LINKED ISSUE #1035: Typescript registration check should consider more factors
The typescript check in `executors/Node` around line 785 checks for ts files in the plugins and suites lists. It should also check functionalSuites since those could be TS as well.

```ts
      if (!require.extensions['.ts']) {
        if (
          ((config.node &&
            config.node.suites.some(pattern => pattern.endsWith('.ts'))) ||
         // check functionalSuites, too
            (config.plugins &&
              config.plugins.some(plugin => plugin.script.endsWith('.ts')))) &&
          typeof this.config.node.tsconfig === 'undefined'
        ) {
          register();
        } else if (this.config.node.tsconfig) {
          register({ project: this.config.node.tsconfig });
        }
      }
```

devpaul:
resolves #1035 

Checks for existence of `tsconfig.json` and if ts file exist in `functionalSuites`

### DIFF
--- a/src/core/lib/executors/Node.ts
+++ b/src/core/lib/executors/Node.ts
@@ -1,58 +1,55 @@
 import { existsSync, readFileSync } from 'fs';
-import { dirname, normalize, resolve, sep, join } from 'path';
-import { sync as nodeResolve } from 'resolve';
 import { CoverageMap, createCoverageMap } from 'istanbul-lib-coverage';
-import { createInstrumenter, Instrumenter } from 'istanbul-lib-instrument';
-import { createSourceMapStore, MapStore } from 'istanbul-lib-source-maps';
 import {
-  hookRunInThisContext,
   hookRequire,
+  hookRunInThisContext,
   unhookRunInThisContext
 } from 'istanbul-lib-hook';
+import { createInstrumenter, Instrumenter } from 'istanbul-lib-instrument';
+import { createSourceMapStore, MapStore } from 'istanbul-lib-source-maps';
+import { dirname, join, normalize, resolve, sep } from 'path';
+import { sync as nodeResolve } from 'resolve';
 import { register } from 'ts-node';
-import { global, deepMixin } from '../../../common';
-import Command from '../../../webdriver/Command';
-import LeadfootServer from '../../../webdriver/Server';
-import Tunnel, { DownloadProgressEvent } from '../../../tunnels/Tunnel';
-
+import { deepMixin, global } from '../../../common';
+import BrowserStackTunnel, {
+  BrowserStackOptions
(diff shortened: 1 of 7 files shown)


src/core/lib/executors/Node.ts
```diff
@@ -782,16 +782,20 @@ export default class Node extends Executor<NodeEvents, Config, NodePlugins> {
...
           register({ project: this.config.node.tsconfig });
+        } else if (typeof this.config.node.tsconfig === 'undefined') {
+          // auto-configure typescript support if we detect its presence
+          if (
+            (config.node &&
+              config.node.suites.some(pattern => pattern.endsWith('.ts'))) ||
+            config.functionalSuites.some(pattern => pattern.endsWith('.ts')) ||
+            (config.plugins &&
+              config.plugins.some(plugin => plugin.script.endsWith('.ts'))) ||
+            existsSync(join(config.basePath, 'tsconfig.json'))
```
jason0x43: We should handle this case (the existence of `<config.basePath>/tsconfig.json`) separately since `config.basePath` may not be the project root. In that case, we would need to pass a `project` to the register command.

src/core/lib/executors/Node.ts
```diff
@@ -782,16 +782,20 @@ export default class Node extends Executor<NodeEvents, Config, NodePlugins> {
...
-        if (
-          ((config.node &&
-            config.node.suites.some(pattern => pattern.endsWith('.ts'))) ||
-            (config.plugins &&
-              config.plugins.some(plugin => plugin.script.endsWith('.ts')))) &&
-          typeof this.config.node.tsconfig === 'undefined'
-        ) {
-          register();
-        } else if (this.config.node.tsconfig) {
+        if (this.config.node.tsconfig) {
```
jason0x43: Minor consistency nit -- is there a reason we're using `this.config` here, but just `config` below?

src/core/lib/executors/Node.ts
```diff
@@ -798,16 +798,25 @@ export default class Node extends Executor<NodeEvents, Config, NodePlugins> {
...
       if (!require.extensions['.ts']) {
-        if (
-          ((config.node &&
-            config.node.suites.some(pattern => pattern.endsWith('.ts'))) ||
+        if (config.node.tsconfig) {
+          register({ project: config.node.tsconfig });
+        } else if (typeof config.node.tsconfig === 'undefined') {
+          // auto-configure typescript support if we detect its presence
+          if (
+            (config.node &&
```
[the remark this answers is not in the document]
devpaul: The `config.node` check comes from `master`, according to the types we should be able to remove this.

Issue #1104, opened 2020-02-17
Method name warning in object interface should be an actual warning


jason0x43:
In `core/lib/interfaces/object`, a warning is logged when the user creates a test with a lifecycle method name. This warning should be emitted as an actual warning with `executor.emit('warning' , ...)` so that it can be displayed outside of a debug run.

PR #1116, opened 2020-02-29
Updates warning log in object to use executor.emit()

samends:
This updates the warning log in the `core/lib/interfaces/object` file to use `executer.emit`

resolves: #1104

### DIFF
--- a/src/core/lib/interfaces/object.ts
+++ b/src/core/lib/interfaces/object.ts
@@ -162,7 +162,7 @@ export function createSuite<S extends typeof Suite, T extends typeof Test>(
         name === 'afterEach'
       ) {
         parent.executor.emit('warning',
-          `Warning: created test with lifecycle method name "${name}"`
+          `Created test with lifecycle method name "${name}"`
         );
       }
 
--- a/src/core/lib/interfaces/object.ts
+++ b/src/core/lib/interfaces/object.ts
@@ -161,7 +161,7 @@ export function createSuite<S extends typeof Suite, T extends typeof Test>(
         name === 'beforeEach' ||
         name === 'afterEach'
       ) {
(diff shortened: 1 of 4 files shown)


tests/unit/tunnels/SeleniumTunnel.ts
```diff
@@ -71,7 +71,7 @@ registerSuite('tunnels/SeleniumTunnel', {
       const oldLog = console.log;
 
       return {
-        afterEach() {
+        afterEachSeleniumTunnelTest() {
```
jason0x43: Hah, the warning caught something!

Since we still want the `afterEach` to run, the fix here should actually be to move the tests into a `tests` object, like:

```js
return {
  afterEach() { ... },
  tests: {
    '3.0.0': createTest(...),
    '3.5.0': createTest(...),
    ...
  }
}
```

src/core/lib/interfaces/object.ts
```diff
@@ -161,7 +161,7 @@ export function createSuite<S extends typeof Suite, T extends typeof Test>(
         name === 'beforeEach' ||
         name === 'afterEach'
       ) {
-        parent.executor.log(
+        parent.executor.emit('warning',
           `Warning: created test with lifecycle method name "${name}"`
```
jason0x43: Minor nitpick -- let's drop the 'Warning:' since the reporter will already prepend 'WARNING' to a warning message.

tests/unit/tunnels/SeleniumTunnel.ts
samends: Tried this code and I'm still seeing the afterEach warning message 🤔

samends:
@jason0x43 updated per your comments, try reviewing again

PR #1117, opened 2020-02-29
Updated browserVersion and version to behave the same

### LINKED ISSUE #1113: Different behavior for `browserVersion` and `version`
In Intern master, `browserVersion` and `version` are being handled differently, at least for MS Edge Chromium.

To test, create a test config with

```
"environments": {"browserName": "MicrosoftEdge", "version": "insider preview" }
```

Run `npm test -- --config <your config> --show-config` and note that the drivers list contains MicrosoftEdge.

Create or update a config to use `"browserVersion": "insider preview"`, run the test command, and note that the drivers list contains MicrosoftEdgeChromium (which is what it's supposed to contain).

samends:
When version is set to `insider preview` and browserName is `MicrosoftEdge`, the driver name should be `MicrosoftEdgeChromium`

resolves: #1113

### DIFF
--- a/src/core/lib/executors/Node.ts
+++ b/src/core/lib/executors/Node.ts
@@ -919,12 +919,14 @@ export default class Node extends Executor<NodeEvents, Config, NodePlugins> {
       ) {
         driverNames.add(browserName);
       } else if (browserName === 'MicrosoftEdge') {
-        const { browserVersion } = env;
+        const { browserVersion, version } = env;
         if (
           (!isNaN(browserVersion) && Number(browserVersion) < 1000) ||
+          (!isNaN(version) && Number(version) < 1000) ||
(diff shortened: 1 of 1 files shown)


jason0x43:
This resolves the edge/edge chromium issue, but it doesn't address the BrowserStack issue. 

Ideally we should handle the problem at an earlier point, such as when the environments are originally read in and normalized, so that `version` and `browserVersion` are always the same.

jason0x43 (changes requested):
This does handle the Edge case, but we should try to deal with this at an earlier point if possible, possibly in `src/core/lib/common/util#processOption`, to cover that and the BrowserStack case.

samends:
Hey Jason, sorry I'm new to this project, what is the `processOption` method used for?

PR #1125, opened 2020-03-31
fix: support timers getting stubbed

jonnycornwell:
- If using a library to stub timers it interferes with the timing
functions used by Suites and Tests
- Tests were written using TDD

### DIFF
--- a/src/core/lib/Suite.ts
+++ b/src/core/lib/Suite.ts
@@ -10,6 +10,7 @@ import Test, { isSkip, isTest, SKIP } from './Test';
 import { InternError } from './types';
 import { Remote } from './executors/Node';
 import { errorToJSON } from './common/util';
+import { setTimeout, clearTimeout, now } from './common/time';
 
 /**
  * The Suite class manages a group of tests.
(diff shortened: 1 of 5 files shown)


jonnycornwell:
Not sure what went on there in the build, did run the tests on node, chrome and firefox locally. Also if you want me to remove the `api.json` changes let me know

src/core/lib/common/util.ts
```diff
@@ -1028,3 +1028,10 @@ export function errorToJSON(error?: InternError): InternError | undefined {
     ...(showDiff ? { actual, expected } : {})
   };
 }
+
+export const originalTimers = {
```
jason0x43: How about we export these from a new `lib/common/time` module, with a comment at the top explaining why we need it? Also, rather than dateNow, how about just `now`? Then we could do

```
import { setTimeout, now } from './common/time';
```

jonnycornwell: That is much nicer and changes have been made

jonnycornwell:
@jason0x43 Is there any chance of getting this change into the next 4.8.X release?

jason0x43:
Yes, that should be doable

jonnycornwell:
Great, thanks!

PR #1126, opened 2020-04-03
docs: Fix simple typo, wihtout -> without

timgates42:
There is a small typo in docs/how_to.md.

Should read `without` rather than `wihtout`.


jason0x43:
Thanks for catching that!

PR #1155, opened 2020-05-19
Webdriver list

### LINKED ISSUE #1071: Create a frequently updated webdriver list
The most common use case with SeleniumTunnel is to use it with current versions of Chrome, Firefox, etc. Default WebDriver versions are hardcoded in SeleniumTunnel and tend to lag behind the current driver releases, which eventually causes problems with the latest browser versions. Updating the default WebDriver versions in SeleniumTunnel and releasing a new Intern just for updated drivers is a bit of a pain. Instead, SeleniumTunnel should use a listing service (we may need to create this; it could just be a file in a repo somewhere, or on Intern's site) that indicates at least the latest version for each driver.

jason0x43:
This PR adds support scripts for updating Intern's webdriver list and publishing the list to Intern's github.io site. It also will update SeleniumTunnel to check for an updated driver list before downloading drivers.

Webdriver versions must still be updated manually (the `webdriver` npm script helps with this)  since none of them but chromedriver have any sort of webservice that indicates the latest stable version.

resolves #1071

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ _examples/
 _tests/
 _testIntern/
 _publish/
+_ghpages/
 src/**/*.js
 selenium-standalone/
 benchmark.json
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -64,6 +64,9 @@ your `intern.json` file that points to the schema, like:
 }
 ```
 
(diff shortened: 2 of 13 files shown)


jason0x43:
This will be backported to Intern 4.8

PR #1176, opened 2020-10-23
Add configuration for websocket timeout

rhpijnacker:
There are cases where it is useful/necessary to set the timeout on acknowledgments from the server when data is sent over the WebSocket. 
This PR makes it possible to configure this value.


src/lib/channels/WebSocket.ts
```diff
@@ -18,10 +18,7 @@ export default class WebSocketChannel extends BaseChannel {
 
   constructor(options: ChannelOptions) {
     super(options);
-
-    if (this.timeout == null) {
-      this.timeout = 10000;
-    }
+    this.timeout = options.timeout || 10000;
```
jason0x43, line 21: Prefer `??` over `||`.

jason0x43 (changes requested):
This is looking good! A couple of the files ended up with Windows line endings -- could you convert those back to `\n`? (They make the diff huge.) Also, could you add the new socketTimeout option to the `schema in src/schemas/config.json`?

jason0x43:
Thanks for fixing the schema and line endings.

Is there a need to use `||` in `this.timeout = options.timeout || 10000;` rather than `??`?

rhpijnacker:
When I use `??` it does not compile anymore.

jason0x43:
Hmm...what version of TS are you building with (`??` is supported by TS 3.7+), and what error does it give you?

rhpijnacker:
Looks like I'm on "Version 3.5.2" (after doing a fresh `npm ci`).

The error is:
```
[0] 5:21:12 PM - [tsc:.] src/lib/channels/WebSocket.ts(21,37): error TS1109: Expression expected.
[0] 5:21:12 PM - [tsc:.] src/lib/channels/WebSocket.ts(21,44): error TS1005: ':' expected.
[0] 5:21:12 PM - Error running tsc:., exiting...
```

jason0x43:
Hmmm...yeah, the TS version is set by intern-dev. Let's restore the original logic, then. While `??` is equivalent to the original, `||` is not (it won't allow a timeout of 0, for example).

jason0x43 (approved):

PR #1180, opened 2021-05-21
Try to bind to different ports if the configured port is already in use

afrischknecht:
Just using a statically configured port is not good enough in some scenarios, e.g. when trying to run multiple intern instances on the same machine.
This PR makes intern try incremental port numbers up to a configured limit in an attempt to discover an unbound port.


afrischknecht:
Sorry, was meant for my fork of the repo.

PR #1185, opened 2021-11-03
Fix circular references

rhpijnacker:
Fixes situations where an AssertionError is sent that contains circular references.


jason0x43:
What about expanding the [stringify](https://github.com/theintern/intern/blob/master/src/lib/common/util.ts#L837) function in `common/lib/util` to handle Errors and then using it?

rhpijnacker:
As suggested on Gitter, reworked it to a more generic solution, using [this MDN snippet](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Cyclic_object_value#circular_references) as the basis.

jason0x43:
Please retarget the PR against master (the monorepo conversion was put on hold, and dev switched back to that). Also, we'll probably need to import a core-js shim (`src/lib/browser/shim.ts`) for WeakMap since IE doesn't support that.

rhpijnacker:
Replaced with #1188 .

PR #1187, opened 2021-11-15
Fix circular references 2

rhpijnacker:
Fixes situations where an AssertionError is sent that contains circular references.
See also #1185.


rhpijnacker:
Too many unrelated changes :(

PR #1189, opened 2022-01-10
Update README.md

veenadevi:
Adding LambdaTestCloud Platform Support


jason0x43:
Intern doesn't have any built-in support for LambdaTestCloud (at least not to the degree that it supports the other mentioned services). If we mention LambdaTestCloud, there should also be a note somewhere about how to use Intern with it.

PR #1207, opened 2023-01-10
Support debug logging (#1150)

### LINKED ISSUE #1150: Support debug logging in browser console output
The Console reporter used when running browser tests doesn't currently (as of Intern 4.8.x / 5.0.0-pre) output debug log statements when `debug` is active. It should.

abolfazlghalandary:
Closes #1150 
I'm solving this issue as a "good first issue". Since I have little experience on frontend, and this is one of my first contributions, I appreciate any step-by-step help to make this PR applicable for merging.

Thanks!

### DIFF
--- a/src/lib/reporters/Console.ts
+++ b/src/lib/reporters/Console.ts
@@ -8,11 +8,17 @@ import Reporter, { eventHandler, ReporterOptions } from './Reporter';
  */
 export default class ConsoleReporter extends Reporter {
   private _hasGrouping: boolean;
+  private _debugLogs: boolean;
   private _testId: keyof Test;
 
-  constructor(executor: Executor, options: ReporterOptions = {}) {
+  constructor(
+    executor: Executor,
+    options: ReporterOptions = {},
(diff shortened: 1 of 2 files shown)


src/lib/reporters/Console.ts
```diff
@@ -22,6 +28,13 @@ export default class ConsoleReporter extends Reporter {
     this.console.error(this.formatError(error));
   }
 
+  @eventHandler()
+  log(data: any) {
+    if (this._debugLogs) {
```
jason0x43, line 33: The reporter doesn't actually need to check any sort of debug condition. Other parts of Intern handle that (see `log` in `src/lib/executors/Executor.ts`), so the Console reporter should simply need to implement a handler for the `log` event.

jason0x43 (changes requested):
Hi, and thanks for contributing! You should be able to simplify this one a bit. The reporter itself doesn't need to worry about whether debug mode is enabled, it just needs to handle `log` events.

